PR #21, opened 2017-01-07
Catch String Errors

lehuff:
* Removed unused import
* Added -se/--stringerror switch, use it if you want to dump the raw page because one of your account strings isn't working correctly. It will create a file for each site with an error it encounters.
* -se/--stringerror is disabled by default


WebBreacher:
Hey Lane. I like this idea and how it works. One small request before I merge: can you change the output file names to something standard like "se-SITENAME" or something so that when you dir/ls the directory after the script executes, all the SE errors are stacked together?

lehuff:
Ok, should be good now.

PR #25, opened 2017-01-08
Check specified sites

L0r3m1p5um:
added a command line option to check a specific site or list of sites


L0r3m1p5um:
Ah, I think this is because I needed to change the line endings in order for the script to run properly on Linux. I'll try to work out how to fix this tomorrow but I'm pretty confident the only changes I made were at lines 45 and 142-153.

L0r3m1p5um:
Ok, I think this is fixed

PR #36, opened 2017-08-22
Refactor code as Python Package

andydennis:
Hey there,

I did a little code refactor and converted this into a Python package. The script can be installed by cloning the repo and using pip, or if we submit it to PyPi by doing

pip install WhatsMyname

Let me know what you think of the idea and the refactor.

Thanks


andydennis:
Oh one more thought. The projects that depend in this may need to update their URLs to pull the JSON if you want to keep it in the src dir.

WebBreacher:
Thanks for taking the initiative and doing this Andy. 

I'm wondering what was the problem that you were solving with this? I try to keep my projects simple and, git clone then run seemed like it hit that mark. I'm curious though. What are your thoughts? Will it reach a larger user-base or....?

Thanks again for the work!

andydennis:
No problem.

I think packaging and adding it to PyPi adds a number of benefits:

1. You can release versions of it going forward for example a python 2.x versus python 3.x. Users can then simply download the version they want directly from pip without having to check the code out. The package will just be added to their environment (of course you can still clone if you want)

2. If another project wants to reuse the classes in WhatsMyName they can add the project as a dependency and then import the classes they wish. With a bit more of a code refactor the arg parse code can be moved out of the main class and params then passed into the class at initialization.   This would allow another project to inherit your class and then provide its own source for inputs e.g name or another JSON file.

3. The setup.py allows you to pin specific versions of dependencies so you don't get into the problem with a new version of a package being released that isn't backwards compatible and is installed by mistake. Having said that we should pin the dependency in the setup.py I added :)

I think as a package it's more flexible in the long run and from an engineering perspective just a little nicer to work with.

Also on the code clean up I reduced some of the duplication and modularized it a bit. This should be easier for maintenance.

Thanks

Andy

WebBreacher:
Andy, thanks again for all the work you put into this PR. After thinking about where I want this project to go and how I want it used, I think I'd like to keep it simple and not make it a python module. The heart of the project is really the JSON file. Making this a module IMO would change the focus. My primary goal is that JSON file (which other tools rely upon).

For those reasons, I'm going to reject this PR. 

I understand you made other changes to the script (fixing spelling and some output) which I would love to include in the project. If you want to submit another PR for just those edits, I'm happy to merge them.

Thanks again for doing all this work. Really appreciate it.

andydennis:
Hey,

Thanks for the feedback. I'll create a separate submit for the other fixes.
If it is alright with you, I plan to maintain fork of the project for those
interested in using it as a package.

Cheers

WebBreacher:
Works for me!

PR #51, opened 2019-02-22
add 'www.' to Instagram-uri

janbinx:
The reason for this being that Instagram returns status code 301 without it. This causes recon-ng to overlook hits as it sets redirect=False.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -685,7 +685,7 @@
       },
       {
          "name" : "Instagram",
-         "check_uri" : "https://instagram.com/{account}/",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Great catch. They must've changed it recently. Thanks for the PR too!

janbinx:
Sure. BTW were you aware that one should use the results from the openstreetmap with caution? In case you have a dot in the username everything on the right side seems to be ignored. openstreetmap.com/users/anthony yields the same result as openstreetmap.com/users/anthony.fsdfjksldfjskldfjslkfd...

PR #89, opened 2019-05-31
Adding Slack

Sector035:


web_accounts_list.json
```diff
@@ -1440,6 +1440,18 @@
          "category" : "travel",
          "valid" : true
       },
+      {
+         "name" : "Slack",
+         "check_uri" : "https://{account}.slack.com",
```
WebBreacher: I don't think this would look for a user named "account" but more a slack group with that name.

Sector035:
That is correct, but there are still a lot of people with a personal Slack.
Looking for actual usernames within the realm of Slack, is impossible sadly
enough.

Give it a test run maybe over time and if you think it is still useful
merge it, else drop it if you don't see the value. 👍🏻

WebBreacher:
I'd prefer to not look for a Slack group with a user's name and, instead, keep focus on individual user account enumeration.

PR #90, opened 2019-05-31
Adding Behance

Sector035:


web_accounts_list.json
```diff
@@ -105,6 +105,17 @@
          "category" : "business",
          "valid" : true
       },
+      {
+         "name" : "Behance",
+         "check_uri" : "https://www.behance.net/{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : "<title>Behance</title>",
```
WebBreacher: When I look at Alice's profile in the site, I get `<title>Alice Wu on Behance</title>` instead of the title string noted above. Can you check it please and adjust?

Sector035: You will see that in the source code of the page after it is loaded, and the browser has had the chance to run all the scripts, performed redirects et cetera. If you simply curl the initial page, you will see the difference. Tested it, it is 100% correct.

WebBreacher: I see what you mean....does the python checker also see the site like the curl request does?

Sector035:
Not sure, but I do know that it broke a few checks I had in mind. There are
sites where the initial load is similar to existing and non existing
accounts, and I can't seem to write a check for that. So I don't think that
it actually follows redirects or dynamically created content.

bcoles:
> 
> Not sure, but I do know that it broke a few checks I had in mind. There are sites where the initial load is similar to existing and non existing accounts, and I can't seem to write a check for that. So I don't think that it actually follows redirects or dynamically created content.

`web_accounts_list_checker.py` uses the Python `requests` library. It does follow redirects, but will not execute dynamic client-side code, such as JavaScript, without the use of a headless browser or JavaScript rendering engine.

For testing and debugging purposes, you can add the following code at line 187 to print the response text and status:

```
        print(r.content)
        print("Status: %s" % r.status_code)
```

In this instance, the majority of the page content is loaded dynamically using JavaScript. This isn't due to requesting the URL with Python `requests` lib (or `curl`), as the same HTML is returned in a browser.

The matches defined in this PR work as described.

```
$ ./web_accounts_list_checker.py -s "Behance" -u doesnotexist | grep "<title>"
<title>Behance :: Oops! We can’t find that page.</title>
$ ./web_accounts_list_checker.py -s "Behance" -u alice | grep "<title>"
<title>Behance</title>
```

I had a quick look over the page source and didn't find a better match. Fortunately, the home page title (`<title>Behance :: Best of Behance</title>`) does not match the above match.

PR #116, opened 2019-08-29
Add social network F3

janbinx:
F3 is an app - like Ask.fm - where you ask people questions anonymously.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -74,7 +74,7 @@
       },
       {
          "name" : "ask.fm",
-         "check_uri" : "http://ask.fm/{account}",
+         "check_uri" : "https://ask.fm/{account}",
          "account_existence_code" : "200",
          "account_existence_string" : "answers,",
          "account_missing_string" : "Page not found",
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -505,6 +505,17 @@
          "valid" : true
       },
       {
+         "name" : "F3",
+         "check_uri" : "https://f3.cool/{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : "Send me anonymous messages or ask anything - F3",
+         "account_missing_string" : "Page Not Found - F3",
+         "account_missing_code" : "404",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -1748,6 +1759,17 @@
          "category" : "news",
          "valid" : true
       },
+      {
+         "name" : "Tik Tok",
+         "check_uri" : "https://www.tiktok.com/@{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : " Following",
+         "account_missing_string" : "Couldn&#x27;t find this account",
```
WebBreacher: Can you check this account_missing_string? Tried view-source:https://www.tiktok.com/@jennifer55555 and cannot find the literal string. The string could be found with a regular ```'```. Also, I found that string in a valid user account HTML.

web_accounts_list.json
```diff
@@ -1748,6 +1759,17 @@
          "category" : "news",
          "valid" : true
       },
+      {
+         "name" : "Tik Tok",
+         "check_uri" : "https://www.tiktok.com/@{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : " Following",
```
WebBreacher: Can you find a better string for the account_existence_string? Something more direct to having a user profile.

janbinx: This is directly linked to having an account. And the json already uses the same method on instagram, periscope, delicious and smule ...

janbinx: And with regards to the other issue: Test it using Python instead of a browser. This should check out as proposed. It does in my end. Your browser must be translating the html entity. Remember a user agent - else you'll get json and not the html.

web_accounts_list.json
janbinx: Test it using Python instead of a browser. This should check out as proposed. It does in my end. Your browser must be translating the html entity. Remember a user agent - else you'll get json and not the html.

WebBreacher:
Thanks for the replies and for submitting this to the project!

PR #119, opened 2019-09-26
web_accounts_list.json file update

decoxviii:
1. kiwibox.com: the site is no longer available
2. SEOClerks: Correct `account_existence_string` returns false/positive

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -1522,7 +1522,7 @@
          "check_uri" : "https://www.seoclerks.com/user/{account}",
          "account_existence_code" : "200",
          "account_existence_string" : "<div class=\"user-info container\">",
-         "account_missing_string" : "User not found",
+         "account_missing_string" : "<title>SEO Marketplace",
          "account_missing_code" : "200",
          "known_accounts" : ["Vfmseo","gokudadon"],
          "category" : "social",
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -1521,7 +1521,7 @@
          "name" : "SEOClerks",
          "check_uri" : "https://www.seoclerks.com/user/{account}",
          "account_existence_code" : "200",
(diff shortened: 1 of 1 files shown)


bcoles:
Confirmed.

Kiwibox no longer exists.

For SEOClerks, the `account_existence_string` string of ` - SEOClerks</title>` will match false positives.

The proposed `account_existence_string` string of `<div class=\"user-info container\">` in the PR is appropriate.

However, the `account_missing_string` string should also be updated. Requests for a non-existent user redirect to the home page. I suggest using `<title>SEO Marketplace` or `<!-- Product List-->` or `Showing page:`.

web_accounts_list.json
```diff
@@ -1532,7 +1521,7 @@
          "name" : "SEOClerks",
          "check_uri" : "https://www.seoclerks.com/user/{account}",
          "account_existence_code" : "200",
-         "account_existence_string" : " - SEOClerks</title>",
+         "account_existence_string" : "<div class=\"user-info container\">",
          "account_missing_string" : "User not found",
```
WebBreacher: Please take @bcoles suggestion on making this account_missing_string more unique. 

> However, the account_missing_string string should also be updated. Requests for a non-existent user redirect to the home page. I suggest using <title>SEO Marketplace or  or Showing page:.

web_accounts_list.json
```diff
@@ -1532,8 +1521,8 @@
          "name" : "SEOClerks",
          "check_uri" : "https://www.seoclerks.com/user/{account}",
          "account_existence_code" : "200",
-         "account_existence_string" : " - SEOClerks</title>",
-         "account_missing_string" : "User not found",
+         "account_existence_string" : "<div class=\"user-info container\">",
+         "account_missing_string" : "<title>SEO Marketplace",
```
decoxviii: The following correction has been made as suggested by @bcoles

PR #148, opened 2020-01-29
Add files via upload

qpointConsulting:
goFindThem.py is a replacement for your web_accounts_list_checker.py in case I royally screwed it up lol. MainFrame is the main GUI window which holds a panel, named UName.py. Done this way for ease of adding more widgets or modifying the view. I run Windows 7 Pro and this works under Python 3.7 - If the main Window gives the famous (Not Responding) ignore it. It will update itself, unless it lasts longer than a minute. wxPython seems to be famous for this and if I ever figure out a way to convince the two to play nice, I'd be very happy!  I use Eclipse IDE for coding as I love their debugger for Python. Thanks!


WebBreacher:
Thanks for this submission. As we discussed offline, since this is a front-end to the WhatsMyName project, I'd prefer to not bundle it with the actual project. Instead, once it is coded to your liking, I'll be happy to link to it to give users a choice on how to use this project's content.

qpointConsulting:
Thanks!

Have a great week

B

PR #155, opened 2020-02-29
Code clean-up

mccartney:
This PR is pure refactoring - i.e. no intended changes in the behavior. Only some minor code clean-up applied:
- first commit is about simplifying the `if something: True else: False` to `something`
- extracting `random_string()` to a function of its own to better readability
- I ran `pylint` static analysis tool against the code and fixed some of the most appealing mistakes reported there


WebBreacher:
thanks for fixing this up

PR #162, opened 2020-03-03
Suggest edits

salaheldinaz:
**Suggest edits:**

- Update web_accounts_list.json
- Create .gitignore
- Add header image to Readme
- Edit Readme file
- Edit check_os() to add 'posix' as default if failed to detect the operating system
- Add descriptions to functions
- Rename finaloutput function to final_output
- fix uname for known_accounts
- Add threading to speed lookup
- edit error style and text
- Save results to file


web_accounts_list_checker.py
```diff
@@ -247,16 +190,136 @@ def finaloutput():
...
+    sites_not_found = len(args.site) - len(data['sites'])
+    if sites_not_found:
+        print(' -  WARNING: %d requested sites were not found in the list' % sites_not_found)
+    print(' -  Checking %d sites' % len(data['sites']))
+
+else:
+    print(' -  %s sites found in file.' % len(data['sites']))
+
+# create result file
+result_file = open("result.txt", "w")
```
mccartney: It seems to me there's no need to introduce a (temporary) output file for just collecting the outcomes from different threads.

WebBreacher:
@salaheldinaz - any way you can submit these as multiple PRs? It'd be easier to accept the simple ones while we discuss the more challenging/bigger changes.

salaheldinaz:
@WebBreacher Ok I will check how I can do that

salaheldinaz:
I created new pull requests with two edits:
- Readme.md file
- Threading for check_online_presence.py

The other edits already implemented (by you) in the last version. ;)

web_accounts_list_checker.py
salaheldinaz: I saw this TODO 
```
2 - Make it so the script will append comment to the entry and output to file
```
that why I added this feature.

@WebBreacher what do you think? Should I add this or not ?

* I will create a new pull if you agree to add it .

PR #166, opened 2020-03-03
Fixing sites en masse

mccartney:
## Context

When working with #164, I noticed tens of sites definitions are outdated:
- the URLs didn't use `https://` while the sites enforce that with a redirect
- site's `missing_string` or `missing_code` were not true
- other minor deviations

## This PR:

- disables redirects as #164 does
- fixes a lot of sites definitions
- removes a few of them (if they no longer work)
- provides other minor changes to site definitions

has been proven to give significantly fewer errors when running:
```
python3.7 -u ./check_online_presence.py
```

### DIFF
--- a/check_online_presence.py
+++ b/check_online_presence.py
@@ -58,7 +58,7 @@ def signal_handler(*_):
 def web_call(location):
     try:
         # Make web request for that URL, timeout in X secs and don't verify SSL/TLS certs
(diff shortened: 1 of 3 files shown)


WebBreacher:
thank you for these submissions. I'd like to keep track of sites we totally remove on the https://github.com/WebBreacher/WhatsMyName/wiki page. I've added to ones you deleted to that page.

PR #171, opened 2020-03-05
Add threading to speed lookup

salaheldinaz:
and PEP8 Formatting

### DIFF
--- a/check_online_presence.py
+++ b/check_online_presence.py
@@ -8,27 +8,30 @@ import signal
 import string
 import sys
 
+import threading
 import requests
-
 
 DEBUG_MODE = False
 
 # Set HTTP Header info.
 HEADERS = {'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:73.0) Gecko/20100101 Firefox/73.0',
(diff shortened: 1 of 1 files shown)


check_online_presence.py
```diff
@@ -135,30 +181,35 @@ def main():
...
         if not site['valid']:
             warn("[!] Skipping %s - Marked as not valid." % site['name'])
             continue
 
         if args.username:
-            check_site(site, args.username,
-                       if_found     = lambda url: positive("[+] User found at %s" % url),
-                       if_not_found = lambda url: neutral( "[-] User not found at %s" % url),
-                       if_neither   = lambda url: error(   "[!] Error. The check implementation is broken for %s" % url))
+            run_type = 'username'
```
mccartney: Extracting the `message_to_print` method makes the code less readable.
And here we are dealing with potentially non-professional developers, so the readability of the code is a key factor.
Is there a better way to structure that?

salaheldinaz: I will look into it

WebBreacher: Hey @salaheldinaz any progress on this? Not pushing...just checking.

salaheldinaz: > Hey @salaheldinaz any progress on this? Not pushing...just checking.
I missed this notification.
I will continue work on it  👍

PR #179, opened 2020-06-13
Adding Elftown

steve-bate:
This site does a 303 redirect to the user's profile page with no response content. I also needed to modify account_existence_string substring check to allow a match at the beginning of the string. If the substring is not found, `find ` will return -1.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -484,6 +484,17 @@
          "valid" : true
       },
       {
+         "name" : "Elftown",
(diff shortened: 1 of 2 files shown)


steve-bate:
I didn't know if this was the correct way to handle a site with a redirect to the user's profile (since the code explicitly prevents the requests library from following redirects).

WebBreacher:
Looks good. Thanks for the edits!

PR #182, opened 2020-07-21
Add threading to speed lookup 2  

salaheldinaz:
- Added threading in `web_accounts_list_checker_threading.py`.

- Changed the output to list results at the end :

> Screen Shot 2020-07-21 at 4 33 04 AM


salaheldinaz:
For in-progress output formatting I can implement use logging it's much better.

Normal print function due to threading :
> Screen Shot 2020-07-21 at 4 44 11 AM

Using logging module:
> Screen Shot 2020-07-21 at 5 05 54 AM

If you prefer using logging module, you can accept this one and I will create a new pull request with logging module implemented.

salaheldinaz:
We can print all errors at once at the end also same as results.

WebBreacher:
logging module would be better IMO.

PR #187, opened 2020-07-30
Update web_accounts_list.json

BerndDasByte:
Fixed platform slideshare (changed account_existence_string)

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -2129,7 +2129,7 @@
          "name" : "slideshare",
          "check_uri" : "https://www.slideshare.net/{account}",
          "account_existence_code" : "200",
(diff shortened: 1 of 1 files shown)


WebBreacher:
While this detection string works for the `test` account, it does not for others (https://www.slideshare.net/john). Can you please re-examine your detection string and make it work for a broader set of accounts?

BerndDasByte:
Could you recheck that? I tested the detection string for the accounts `john`, `master`, `goku` and `dragon` and found them all.

BerndDasByte:
Have you checked it multiple times? Sometimes there are False Negatives, why ever ...

WebBreacher:
Yes! Apologies for the delay. rechecked on several accounts and it works just fine like you said. THANKS for the submission!

PR #202, opened 2020-09-30
Configured requirments.txt

DanBrown47:


requirements.txt
```diff
@@ -1 +1,5 @@
-requests
+requests==2.24.0
+argparse==1.4.0
```
WebBreacher, line 2: Argparse is part of the Python Standard Library and we don't need a specific version.

Running `pipreqs` on my dev system yielded the following requirements file:
```
urllib3==1.25.8
requests==2.22.0
```

web_accounts_list_checker_threading.py
```diff
@@ -81,7 +81,7 @@
 
 def check_os():
     """
-    # check operating system ot adjust output color formatting
+    # check operating system or adjust output color formatting
```
WebBreacher, line 84: Good catch on this spelling error.

WebBreacher (changes requested):
Can you adjust the requirements.txt file to remove argparse and alter the versions required for the other modules to reflect:

```
urllib3==1.25.8
requests==2.22.0
```

requirements.txt
DanBrown47: I have removed the argparse :+1:

PR #212, opened 2020-10-26
Added fabswingers.com as a site to be checked

swedishmike:
Added fabswingers.com. Tested and it works fine - the only thing I'm a bit uncertain about is the category. It's "dating" but of the more "XXX" kind. ;)

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -5,7 +5,7 @@
                 "http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to",
                 "Creative Commons, PO Box 1866, Mountain View, CA 94042, USA."],
    "authors" : ["WebBreacher","Munchko","L0r3m1p5um","lehuff","janbinx","bcoles","Sector035",
-                "arnydo","mccartney","salaheldinaz","camhoff","jocephus"],
+                "arnydo","mccartney","salaheldinaz","camhoff","jocephus","swedishmike"],
    "categories" : ["art","blog","business","coding","dating","finance","gaming","health",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -2807,6 +2807,17 @@
...
          "valid" : true
+      },
+      {
+         "name" : "Fabswingers",
+         "check_uri" : "https://www.fabswingers.com/profile/{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : "View Profile",
+         "account_missing_string" : "The user you tried to view doesn't seem to be on the site any more",
+         "account_missing_code" : "200",
+         "known_accounts" : ["fabswingers.com"],
```
WebBreacher, line 2818: Would you mind changing this value from `fabswingers.com` (which looks like a fake account because it is a domain) to some other, real accounts like `foreveryours87` and `viciqueen`? Or at least add them to the list of `known_accounts`?

WebBreacher (changes requested):

web_accounts_list.json
swedishmike: This is the admin/owners account so I choose that one since that one wouldn't disappear as easily as a random one I could find.

WebBreacher: Makes sense. Would you mind adding 1 or 2 other "user" accounts?

swedishmike: Sorry - wasn't logged in to GH on the host I made the changes from so couldn't reply but they should be good now.

PR #213, opened 2020-10-27
Fixes #209 + removed Scribd since the config was for scratch.

### LINKED ISSUE #209: Music sites to add
https://www.magix.info/us/users/profile/baywolfmusic/
https://www.tunefind.com/user/profile/baywolfmusic

swedishmike:
@WebBreacher  - sorry, hadn't seen the issue/request for those two sites when I submitted my first one. They seem to work and I've found another profile per site to add as well as 'knowns'.

The Scribd site wasn't correct, it was just the name that said Scribd. The rest was copied from the Scratch entry so removed it.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -2058,17 +2058,6 @@
          "valid" : true
       },
       {
-         "name" : "Scribd",
-         "check_uri" : "https://scratch.mit.edu/users/{account}/",
-         "account_existence_code" : "200",
-         "account_existence_string" : "on Scratch</title>",
-         "account_missing_string" : "We couldn't find the page you're looking for.",
-         "account_missing_code" : "404",
-         "known_accounts" : ["griffpatch"],
-         "category" : "coding",
-         "valid" : true
-      },
-      {
          "name" : "SEOClerks",
          "check_uri" : "https://www.seoclerks.com/user/{account}",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -2818,6 +2807,28 @@
...
+         "category" : "music",
+         "valid" : true
+      },
+      {
+         "name" : "Tunefind",
+         "check_uri" : "https://www.tunefind.com/user/profile/{account}",
+         "account_existence_code" : "200",
+         "account_existence_string" : "Achievements",
+         "account_missing_string" : "Page not found",
+         "account_missing_code" : "200",
```
WebBreacher, line 2828: After testing this, I think this should be a 404. The `python3 ./check_online_presence.py -s tunefind` comes back with a false negative for the random name. Can you check it out?

swedishmike: My bad. I used `web_accounts_list_checker.py` to test with. :(

404 doesn't work, if it's an incorrect profile it 302's to a 404 so the 302 seems to work better. I've tested with `check_online_presence.py` now and that seems to go through without errors now. Will submit updated .json.

WebBreacher: Looks good. Thanks for fixing.

PR #214, opened 2020-10-28
Three new sites added. Appian, Digitalspy and Moneysavingexpert.

swedishmike:
As per the title, Appian, Digitalspy and Moneysavingexpert have been added. 

This time I've checked them with `python3 ./check_online_presence.py -s` and they all come out ok.


WebBreacher:
Thanks for the submissions!

PR #220, opened 2020-10-29
Add Ko-Fi

arnydo:
Adding https://ko-fi.com.
A "Buy me a coffee" site.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -343,6 +343,8 @@
          "name" : "Ko-Fi",
          "check_uri" : "https://ko-fi.com/{account}",
          "account_existence_code" : "200",
+         "account_existence_string" : ">  Buy a Coffee for",
+         "account_missing_string" : "Object moved to",
          "account_missing_code" : "302",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -339,6 +339,15 @@
          "category" : "shopping",
          "valid" : true
       },
+      {
+         "name" : "Ko-Fi",
+         "check_uri" : "https://ko-fi.com/{account}",
+         "account_existence_code" : "200",
```
WebBreacher, line 345: Thanks for the submission. For this entry to be included in the project, I need you to find the `account_existence_string` (something unique to the page where an account is) and the `account_missing_string` (something specific in the response from the server when an invalid account was requested).

WebBreacher:
Thanks for the submission!! Please add the existence and missing strings and resubmit. This is a terrific site to add.

web_accounts_list.json
arnydo: There you go. I left them out initially simply because non-existent profiles always result in 302. I added in the existing and missing strings. Thanks!

Issue #229, opened 2021-01-04
Colour output bit of a hit and miss on non-Linux os's


swedishmike:
I've had varying success with the coloured outpout on MacOS and Windows 10.

This comes down to the os checks and colour definitions in the scripts after what I can gather.

WebBreacher:
I've been thinking about this and, if possible, i'd rather not include this non-core library in the scripts. I get the colorization issue and Win systems. Wonder if there is a different way to address it?

swedishmike:
I fully see your point here... I took the easy way out and used something someone who had done the hard work had made public - I'll have a think and a play around to see what can be done 'more natively' too.

WebBreacher:
I wonder if the issue with the existing script(s) is the detection of Windows systems and not how it makes the colors? Also, what Windows app are you using? In WSL (Windows Subsystem for Linux) terminal, I see the output like the Linux/Mac users do.

swedishmike:
That could very well be it... I think the build I was having to use doesn't have/allow the WSL so it would have been the 'normal' one.

Maybe the easiest way would be add the following to the documentation: "If you want colors in Win10 - use WSL". ;)

PR #230, opened 2021-01-04
Changes made to address the colour output issue #229

swedishmike:
This is changes to address issue #229 that I opened earlier.

These changes are quite invasive, including a new module ( https://github.com/willmcgugan/rich ) that handles the colours automagically for different OS's. I've updated `requirements.txt` to reflect this. Since the other modules are locked to versions I did the same to rich.

I've updated and tested the three different scripts and it looks like it's working as expected on all three of them now, I could of course still have missed testing one or two cases. The most changes were made to `web_accounts_list_checker_threading.py` since that one had been moved to use `logging` rather than print - to get them to work roughly the same I reverted those changes. Something that might be ok/liked or not.

Have a look at what I've done and see if it's something that you think could be good for the project and give a more similar experience across different OS's.

### DIFF
--- a/check_online_presence.py
+++ b/check_online_presence.py
@@ -3,14 +3,13 @@
 import argparse
 import collections
 import json
-import os
 import random
 import signal
 import string
 import sys
 
 import requests
-
+from rich import print
 
 DEBUG_MODE = False
 COUNTER = collections.Counter()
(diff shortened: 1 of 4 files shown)


WebBreacher:
Thanks for the PR @swedishmike ! What is the issue that you are trying to fix here? Was there a problem in the colorization on different systems? I'm looking at the code and it looks very similar and now makes the script rely on an additional library.

If this is just cosmetic, I'm disinclined to accept the PR. Thoughts?

swedishmike:
Hi there...

It was to fix the lack of colorisation on Windows mainly, to make it the same as the other OS's and give the poor Windows users a bit better experience.

I had to run it on Win10 a couple of times and realised that I actually missed the colors a bit.

However, if you feel that getting the same experience across doesn't weigh up another library I fully see your point.

WebBreacher:
Per our discussion in the #230 issue, I'm going to reject/close this PR. If you end up reworking the color to use core modules/libraries, please let me know/resubmit.

And thanks for all your work on the project!!! Appreciate your work @swedishmike !

PR #231, opened 2021-01-11
Added optional input & output file toggles

jspinel:
Argument: -of 
Writes to a specified outputfile, does not require -o to be toggled

Argument: -in
Allows for additional input files to be used, for example if other sites want to be used or if WhatsMyName is called from a different directory.


WebBreacher:
Thank you for the PR!

PR #246, opened 2021-01-21
Fixed Sites

spotlightc:
* Removed: Oracle and Spotify
* Removed: Stunited - found users via googling (site:stunited.org "profile") but was not able to get to the sites
* Added prety_uri to Diigo
* Added new user to eBay
* Changed Wireclub to be true

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -428,17 +428,6 @@
          "valid" : true
       },
       {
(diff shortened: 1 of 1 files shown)


WebBreacher:
Checked and saw no issues. Comment shows `wireclub` was marked as `true` but I think you meant `tamtam`.

PR #253, opened 2021-02-02
Dojoverse - second attempt

swedishmike:
This looks better. Sorry about that - not sure what went wrong, not sure if VS Code was having an episode. Tried with that again and the same behavior, this is just done with Atom and seems to work.

Tests are still working.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -3148,6 +3148,20 @@
          "known_accounts" : ["john", "blue"],
          "category" : "social",
          "valid" : true
+      },
+      {
+         "name": "Dojoverse",
+         "check_uri": "https://dojoverse.com/members/{account}/",
+         "account_existence_code": "200",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -3148,6 +3148,20 @@
...
          "valid" : true
+      },
+      {
+         "name": "Dojoverse",
+         "check_uri": "https://dojoverse.com/members/{account}/",
+         "account_existence_code": "200",
+         "account_existence_string": "Joined",
+         "account_missing_string": "Looks like you got lost!.",
+         "account_missing_code": "404",
+         "known_accounts": [
```
WebBreacher, line 3159: can you place this list of accounts on the same line like in line 3148 above?

swedishmike:
Should be ok now .

WebBreacher:
Thanks for the edits!

PR #260, opened 2021-03-03
Removed Hacker News dup

steve-bate:
There were two entries for the Hacker News website (named "Hacker News" and "Hackernews").


WebBreacher:
thank you for catching this

PR #271, opened 2021-04-24
Removed plug.dj and massRoots from web_accounts_list

tomgold182:
plug.dj and massroots are no longer exists and the module is crashing when tries to search them.
I have removed both of the site from the web_accounts_list json file until I will have more time to fix the catch phrase so it won't lead to a crash and move the next site if reaching max retries

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -430,17 +430,6 @@
          "valid" : false
       },
       {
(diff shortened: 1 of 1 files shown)


WebBreacher:
Thanks for catching those and submitting a PR.

PR #289, opened 2021-06-28
Fixed false positive error with Yahoo Auction

ef1500:
A false positive would show due to Yahoo Auction Profile pages sending 200 codes even when an account did not exist.


WebBreacher:
Thank you for finding and fixing this.

PR #291, opened 2021-08-07
Use Duolingo's api directly

ef1500:
There was a small false negative error when trying to enumerate a duolingo account because it takes a few seconds for the webpage to load. This Takes Care of it.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -234,6 +234,7 @@
       {
         "name" : "Duolingo",
         "check_uri" : "https://www.duolingo.com/2017-06-30/users?username={account}&_=1628308619574",
+        "pretty_uri" : "https://www.duolingo.com/profile/{account}",
         "account_existence_code" : "200",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Thanks for catching this and for taking the time to research and make this PR. Could you make 1 change please? Create the `pretty_uri` parameter with a value of `"https://www.duolingo.com/profile/{account}"`. Some of the downstream apps that use WhatsMyName check for the acct using the `check_uri` and then display the results using the `pretty_uri`.

WebBreacher:
Here is an example from the Gab entry:
```
        "name" : "Gab",
        "check_uri" : "https://gab.com/api/v1/account_by_username/{account}",
        "pretty_uri" : "https://gab.com/{account}",
```

ef1500:
Sure thing. Should be good to go now!

WebBreacher:
Thanks!

Issue #293, opened 2021-08-18
Check garmin and protonmail for false positives


WebBreacher:

WebBreacher:
### Protonmail

Protonmail's api has changed so that ANY email (valid or not) that is provided to https://api.protonmail.ch/pks/lookup?op=index&search=adasdafef43@protonmail.com will show positive results (i.e., there is an email there) instead of positive for some and negative for others. WhatsMyName was using this to find protonmail accounts and not it cannot discern truth from false positive.

I've poked the API and found the public key returning URL https://api.protonmail.ch/pks/lookup?op=get&search=webbreacheeegergergr@protonmail.com but that too provides positive data no matter what is presented. 

### Garmin

Changed detection string.

See #294

seewhen:
Hi there,

I've been looking at the Proton issue and I've found a new endpoint you can query for valid accounts. 

The URL https://account.protonmail.com/api/users/available?Name=USERNAME is used during the sign-up phase for new accounts. A get request can be sent to this endpoint along with a username and the site will return a code as JSON indicating whether or not the username is in use or not. 

**Code: 1000**: (no message returned, so the username is available) 
**Code: 12101**: "Invalid input"
**Code: 12102**: "Username contains invalid characters"
**Code: 12106** :"Username already used"

Unfortunetly this endpoint requires the header "x-pm-appversion" with the value "WebAccount_4.4.1". If not you'll get the error "Missing x-pm-appversion header" when sending the get request. 

I know sending unique headers per request is not possible at the moment, so this comment is for documentation purposes should WMN decide to implement that feature.

WebBreacher:
I'm working with OSINT Combine to implement the http_header value that @mklarz submitted in the #295 PR.

WebBreacher:
Protonmail is now being checked in the hosted app. We changed the format of the http_headers key/value.

PR #295, opened 2021-09-05
Add support for HTTP headers and fix ProtonMail

mklarz:
Adds an optional key (`http_headers`) which can be used to specify additional HTTP headers. This fixes ProtonMail.

Related to issue #293

### DIFF
--- a/check_online_presence.py
+++ b/check_online_presence.py
@@ -57,10 +57,13 @@ def signal_handler(*_):
     error(' !!!  You pressed Ctrl+C. Exiting script.')
     sys.exit(130)
 
-def web_call(location):
+def web_call(location, additional_headers=None):
+    request_headers = HEADERS.copy()
+    if additional_headers:
+        request_headers = {**request_headers, **additional_headers}
     try:
         # Make web request for that URL, timeout in X secs and don't verify SSL/TLS certs
-        return requests.get(location, headers=HEADERS, timeout=60, verify=False, allow_redirects=False)
+        return requests.get(location, headers=request_headers, timeout=60, verify=False, allow_redirects=False)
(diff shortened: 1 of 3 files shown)


WebBreacher:
before I accept this PR, I need to ensure that adding a new name/value pair for `"http_headers" : {"x-pm-appversion": "Other"},` will not have adverse impacts on the WhatsMyName.app site.

mklarz:
Fair enough, I have only been testing it directly through the script, but it should hopefully work as-is.

WebBreacher:
my concern would be that enabling the PM entry will cause false positives since the whatsmyname.app site has not implemented (if it can), the header injection.

mklarz:
ah right. in that case it would indeed result in false positives as it will always fail, and therefore report existing accounts as non-existing. I can fix that for you if you wish, but the src isn't public right? Should only need to iterate through `http_headers` (if they exist) and use `XMLHttpRequest.setRequestHeader()` in that `checkStatus` method.

WebBreacher:
You are correct that it is not Open Source code and I do not control it either. I've asked OSINT Combine (the owner) to consider making these changes.

If you want to break out the changes into 1 PR for the script changes and 1 for the PM JSON changes, I can merge the script changes already.

WebBreacher:
Testing part of this out by manually adding to #304. If that works in the modded web site app, I'll add the rest in.

WebBreacher:
Protonmail is now being checked in the hosted app. We changed the format of the http_headers key/value.  Because of that @mklarz I'd like to ask you close this PR and alter the checker scripts to use the current format: `"http_header_key" : "{x-pm-appversion:Other}",`

PR #329, opened 2021-10-17
Search wayback for archived twitter and parler accounts

jocejocejoe:
Use a couple different methods to search for archived profiles, posts, and tweets.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -3535,6 +3535,54 @@
         "known_accounts": ["Bstar3Dart","Kimbolt"],
         "category": "hobby",
         "valid": true
+      },
+      {
+        "name" : "Parler Profile in Wayback",
+        "check_uri" : "http://archive.org/wayback/available?url=https://parler.com/profile/{account}",
+        "pretty_uri" : "https://web.archive.org/web/2/https://parler.com/profile/{account}",
+        "account_existence_code" : "200",
+        "account_existence_string" : "\"archived_snapshots\": {\"closest\"",
+        "account_missing_string" : "\"archived_snapshots\": {}",
+        "account_missing_code" : "200",
+        "known_accounts" : ["JoePags","dineshdsouza"],
+        "category" : "social",
+        "valid" : true
+      },
+      {
+        "name" : "Parler Posts in Wayback",
+        "check_uri" : "http://archive.org/wayback/available?url=https://parler.com/profile/{account}/posts",
+        "pretty_uri" : "https://web.archive.org/web/2/https://parler.com/profile/{account}/posts",
+        "account_existence_code" : "200",
+        "account_existence_string" : "\"archived_snapshots\": {\"closest\"",
+        "account_missing_string" : "\"archived_snapshots\": {}",
+        "account_missing_code" : "200",
+        "known_accounts" : ["JoePags","dineshdsouza"],
+        "category" : "social",
+        "valid" : true
+      },
+      {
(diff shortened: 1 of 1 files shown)


jocejocejoe:
This is more of an RFE - I'm not sure if this is within scope for the project, or if it's implemented in a nice way, but maybe it's a starting point for discussion.

WebBreacher:
Hey @jocejocejoe . Thanks for the submissions. I think this is a cool idea and definitely within the scope of the tool. Can you:

1. Add `pretty_uri` entries to these submissions so that, when tools like whatsmyname.app find positive hits, they direct users to the actual archive.org URL for the archived web page instead of the API? Look at the `Gab` or other entries already in the JSON for examples.
2. Resubmit the pull request with the conflict resolved? Looks like, because I merged your previous PR before this one, there is now a conflict in the JSON file.

jocejocejoe:
Actually, @WebBreacher I'd appreciate your thoughts on how best to handle this wayback archive search idea.

In theory, one could search wayback for almost any account on any site, by looking for the presence of the same URL in the archive, and probably checking for the same strings. But that would be something for consumers of the json file to implement, not something to add to the file manually for each site. This would also allow behavior like "only show archived accounts if it's not currently live" if that was desired.

On the other hand, sites like twitter and parler are heavily archived (and reasonably archive-able).  And sites like Parler in particular are known to have many accounts in the archive which are not present in the the current reincarnation. Other sites like "freedomlake.com" have gone offline permanently but could still be searched in wayback, i.e.

```
      {
        "name" : "Freedom Lake Wayback Machine",
        "check_uri" : "http://archive.org/wayback/available?url=http://freedomlake.com/{account}",
        "pretty_uri" : "https://web.archive.org/web/submit?url=http://freedomlake.com/{account}",
        "account_existence_code" : "200",
        "account_existence_string" : "\"archived_snapshots\": {\"closest\"",
        "account_missing_string" : "\"archived_snapshots\": {}",
        "account_missing_code" : "200",
        "known_accounts" : ["LexTalionis","TheVigilAmerican"],
        "category" : "social",
        "valid" : true
      }
```

Adding wayback searches directly to the json would also mean that there is no work for any client to do; the search is built into the json and results are simply returned like any others.

What do you think might be best? Does it make sense to add specific wayback searches to the json itself, in particular for sites which are still live?

WebBreacher:
Hey @jocejocejoe. Wanted to check if you'd still liike to do this. If so, I'll create the new category.

WebBreacher:
Thank you @iamzewen for resubmitting this as #374 and thank you @jocejocejoe for proposing it and doing all the hard work.

jocejocejoe:
Oh hey @WebBreacher, sorry for missing the discussion and thanks to @iamzewen for fixing up and resubmitting this. I did want to ask how you felt about having ~duplicates in there, i.e. finding archived `https://parler.com/profile/{account}` and/or `https://parler.com/profile/{account}/posts` - might be nicest to somehow combine them, though OTOH it's entirely possible for them to be different sets of information. It all works fine, just seems a little cluttered. Maybe there's no better way, as unique information frequently exists only in one of the two URLs.

WebBreacher:
hmm... if the posts are under the profile, then we really just need to show the profile UNLESS the search of the posts shows when the username is mentioned in someone ELSE's posts. So, @iamzewen tweets to/about "webbreacher". When searching the archive.org site, that Twitter Posts search should show that as well as posts from me. Correct? If so, I'm in favor of leaving it with one of profiles and one for posts.

PR #341, opened 2021-12-28
feat: add argument that enables following redirects

P3run:
When preparing my own list I found that some sites always return 301 or 302 code no matter if the user exists or not. Following redirects fixes that problem as we are redirected to the different pages.


WebBreacher:
Thanks for these enhancements!

PR #344, opened 2022-01-07
feat: add support for optional parameter containing cookies

P3run:
Some pages require certain cookies to be sent with the request. I have added an optional parameter where you can specify such cookies.


WebBreacher:
Hmmm... So how do you picture this working since each cookiue will only work for a given site (that is I cannot use the GitHub cookie on other sites)? A user would need to supply the individual cookies for every site requiring it and we would need to make separate cmdline statements for each. 

Before we change the parameters in the JSON and add new ones, we also need to think about downstream uses of the JSON that are browser-based and could not supply these parameters. 

Can you tell me more about your ideas here?

P3run:
I will give you an example - I wanted to add an adult site to my list. Everything worked fine when I was requesting the profile page in the browser but the script couldn't make it. Then I realized that when I reached that site for the first time I had to confirm that I'm 18+. That site stored that information in the "accept_rules" cookie. Without that cookie I was redirected to the "welcome page". So my solution was as follows:
{
      "name": "showup.tv",
      "check_uri": "https://showup.tv/profile/{account}",
      "account_existence_code": "200",
      "account_existence_string": "Napiwki",
      "account_missing_string": "404",
      "account_missing_code": "404",
      "known_accounts": [
        "Medroxyprogesterone",
        "KatarinaKasya"
      ],
      "category": "XXXPORNXXX",
      "valid": true,
      "cookies": {
        "accept_rules": "true"
      }  
    }
    
The "cookies" parameter is optional (see the if statement in the "web_accounts_list_checker.py") so when the site does not require it we can skip it. The "value" of this parameter is an object containing the cookies that we want to inject to the request (so in fact we can inject many different cookies). So far I found only one site for which I had to inject cookie but I assume that there could be more similar sites out there. 

Does this answer your question?

WebBreacher:
Understood. Unfortunately, I do not think this is something I'd like to add to this project as, in addition to each response code and HTML response, we would need to collect/extract cookies and those cookies could only be used in the Python/non-Web (not https://whatsmyname.app) tools. I'm imagining the vast majority of the users of this project use the web site at whatsmyname.app. Adding sites that required cookies to be submitted would add another layer of complexity that I don't think we want to add.

P3run:
Okay, got it :(. Maybe I will just add it my fork.

soxoj:
For the sake of history, I will note that users could be interested in passing their own authorization cookies (cause for many platforms it allows extracting a lot of additional information from a profile page). I understand that it is too complex and redundant for the current usage of WhatsMyName, just remind you that such a need exists.

WebBreacher:
Yes some websites require the presentation of cookies, HTTP POST instead of GET requests, javascript validation, and other additional data to be presented at the request. Unfortunately, those more complicated requests go beyond the current implementation of WhatsMyName.

Thanks for the submission @P3run....please keep those features coming! Closing this one without including.

PR #347, opened 2022-01-18
add Polish websites

P3run:
I had to remove some of the websites that are present in my list because they work only when the script follows redirects however this commit contains more than 50 sites.

NOTE: The "Joe Monster" site has two queries - one for private and one for public accounts.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -14,7 +14,6 @@
      {
        "name" : "7cup",
        "check_uri" : "https://www.7cups.com/@{account}",
-       "cookies": {"admin": "true"},
        "account_existence_code" : "200",
        "account_existence_string" : "Profile - 7 Cups",
        "account_missing_string" : "Oops! The content you're attempting to access could not be found.",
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -6,14 +6,15 @@
...
-              "jspinel","ef1500","zewen","jocejocejoe"],
+              "jspinel","ef1500","zewen","jocejocejoe", "P3run"],
   "categories" : ["art","blog","business","coding","dating","finance","gaming","health",
                 "hobby","images","misc","music","news","political","search","shopping","social",
                 "tech","video","XXXPORNXXX"],
   "sites" : [
      {
        "name" : "7cup",
        "check_uri" : "https://www.7cups.com/@{account}",
+       "cookies": {"admin": "true"},
```
WebBreacher, line 17: Oops. I think you might have missed removing this `cookies` tag. Can you remove it?

WebBreacher:
I'm wondering if we should tag any of these sites as country-specific? Like `social-pl` or something. Thoughts?

web_accounts_list.json
WebBreacher: I thought so. Really appreciate your additions here.

PR #349, opened 2022-01-18
Fixed Minecraft.list

spotlightc:
Changed known accounts for Minecraft.list


WebBreacher:
closing PR since new one is coming with same content

PR #352, opened 2022-01-19
Add adult sites

P3run:
This time I want to add some non-Polish adult sites.
I also fixed a typo in "skeb".

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -4306,7 +4306,7 @@
           "account_missing_string" : "There is an error",
           "account_missing_code" : "404",
           "known_accounts" : ["xkarola","zorro3330"],
-          "category" : "dating",
+          "category" : "XXXPORNXXX",
           "valid" : true
         },
         {
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -4263,6 +4263,62 @@
       "known_accounts" : ["summit1g","cohhcarnage"],
       "category" : "gaming",
       "valid" : false
-      }
+      },
+      {
+        "name" : "fansly",
+        "check_uri" : "https://apiv2.fansly.com/api/v1/account?usernames={account}",
+        "pretty_uri": "https://fansly.com/{account}/posts",
+        "account_existence_code" : "200",
+        "account_existence_string" : "username",
+        "account_missing_string" : "response: []",
+        "account_missing_code" : "200",
+        "known_accounts" : ["Mikomin","test"],
+        "category" : "XXXPORNXXX",
+        "valid" : true
+        },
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -4263,6 +4263,62 @@
       "known_accounts" : ["summit1g","cohhcarnage"],
       "category" : "gaming",
       "valid" : false
```
WebBreacher, line 4265: Does the pretty_uri change now make this site work? If so, lets change the "valid" to `true`

web_accounts_list.json
```diff
@@ -4263,6 +4263,62 @@
...
+        "known_accounts" : ["Mikomin","test"],
+        "category" : "XXXPORNXXX",
+        "valid" : true
+        },
+        {
+          "name" : "mym.fans",
+          "check_uri" : "https://mym.fans/{account}",
+          "account_existence_code" : "200",
+          "account_existence_string" : "posts",
+          "account_missing_string" : "Il n’y a rien ici…",
```
WebBreacher, line 4284: I think this response is language specific. When I go to the site, my browser shows "There’s nothing here...". Can you pick a different existence_string that is not lang dependent?

web_accounts_list.json
```diff
@@ -4263,6 +4263,62 @@
...
+        },
+        {
+          "name" : "datezone",
+          "check_uri" : "https://www.datezone.com/users/{account}/",
+          "account_existence_code" : "200",
+          "account_existence_string" : "Profile",
+          "account_missing_string" : "There is an error",
+          "account_missing_code" : "404",
+          "known_accounts" : ["xkarola","zorro3330"],
+          "category" : "dating",
```
WebBreacher, line 4309: I know this appears to be a dating site, but looking at some of the blurry profile pictures, those are not safe for work. I can tell what is in those pics. Because of that, please move this to XXXPORNXXX instead of dating.

P3run: Okay, I wasn't sure how to categorize it.

web_accounts_list.json
P3run: Oh, I didn't realize that it was marked as not valid. I just spotted the inconsistency in the parameter name.

web_accounts_list.json
P3run: Have you tried to test this entry by running the script?

web_accounts_list.json
WebBreacher: thank you.

WebBreacher:
Thanks for your time here @P3run !!

Issue #379, opened 2022-01-28
AnonUp (site add request)


jevarlet:
https://anonup.com/@[username]

WebBreacher:
Done! Thanks for the suggestion

jevarlet:
Thank you @WebBreacher!

Issue #378, opened 2022-01-28
WeGo.Social (site add request)


jevarlet:
https://wego.social/[username]

WebBreacher:
Done! Thanks for the suggestion

jevarlet:
Thank you again

PR #380, opened 2022-01-28
Added Sites

spotlightc:
anonup and wego


WebBreacher:
Added per issues #378 and #379

PR #397, opened 2022-04-09
New solution for Twitch.tv

iamzewen:
I've been wanting to switch the URL for Twitch for a while now as the URL we've been using, https://passport.twitch.tv/usernames/USERNAME, is not the prettiest of URLs to link to. The passport site seem also to have had some updates to it which may have broken this specific endpoint as well - time for something new I guess! 

I have added a new URL for our checks, twitchtracker.com, which seem to do the job well. It is used a lot in the Twitch community and is often the go-to site for users when they want stats and general insight into a Twitch account.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -3956,6 +3956,7 @@
     {
       "name" : "Twitch",
       "check_uri" : "https://twitchtracker.com/search?q={account}",
+      "pretty_uri" : "https://twitch.tv/{account}/",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Sounds good and thank you for this work. Can you please use this new url for the check_uri and then add a pretty_uri that points directly to the twitch user account profile? We do that for other platforms where we use a third party site to validate profiles.

PR #399, opened 2022-05-28
✨ adding sites

djahren:
Added a handful of sites, notably Spotify and Snapchat.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -4438,7 +4438,7 @@
           "valid" : true
         },
         {
-          "name" : "Snapchat",
+          "name" : "Snapchat Stories",
           "check_uri" : "https://story.snapchat.com/s/{account}",
           "account_existence_code" : "200",
           "account_existence_string" : "is on Snapchat!",
@@ -4477,7 +4477,7 @@
            "account_existence_string" : "This blog contains adult content",
            "account_missing_string" : "This blog doesn't exist.",
            "account_missing_code" : "200",
-           "known_accounts" : ["themunch", "coffeemeandporn"],
+           "known_accounts" : ["themunch", "shibari4all"],
            "category" : "XXXPORNXXX",
            "valid" : true
           },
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -4436,6 +4436,72 @@
...
+          "valid" : true
+          },
+          {
+           "name" : "BDSMLR",
+           "check_uri" : "https://{account}.bdsmlr.com",
+           "account_existence_code" : "200",
+           "account_existence_string" : "This blog contains adult content",
+           "account_missing_string" : "This blog doesn't exist.",
+           "account_missing_code" : "200",
+           "known_accounts" : ["themunch", "coffeemeandporn"],
```
WebBreacher, line 4480: Can you pick a different account instead of "coffeemeandporn" as that one now requires authentication to view.

web_accounts_list.json
```diff
@@ -4436,6 +4436,72 @@
           "known_accounts" : ["tata23", "fitzsta"],
           "category" : "XXXPORNXXX",
           "valid" : true
-        }
+        },
+        {
+          "name" : "Snapchat",
```
WebBreacher, line 4441: We already have a Snapchat entry. Can you change this to a name of "Snapchat Story" please since it is another valid detection string?

WebBreacher (commented):
Please address these issues and then I can approve this. Thanks for the submission!!

web_accounts_list.json
djahren: Absolutely! I should do a better job of double-checking beforehand but I guess having the alternate detection method isn't a bad thing, eh?

web_accounts_list.json
djahren: Changed to "shibari4all" which doesn't require auth and modified the detect string to work for accounts/blogs that are public and ones that require auth.

djahren:
> Please address these issues and then I can approve this. Thanks for the submission!!

@WebBreacher Made those modifications. Happy to contribute to such a great project. 🙂

PR #401, opened 2022-06-10
Add check for WordPress Blog account

Unpublished:


web_accounts_list.json
```diff
@@ -2443,7 +2443,17 @@
        "valid" : true
      },
      {
-       "name" : "WordPress",
+       "name" : "WordPress Blog",
+       "check_uri" : "https://{account}.wordpress.com",
+       "account_existence_code" : "200",
+       "account_existence_string" : "html",
```
WebBreacher, line 2449: Please choose a more unique `account_existence_code`. Something that only can be found in the positive result.

Unpublished: I intentionally chose this unspecific to match both an activated blog (test) and a reserved blog (unknown31). I'd even have left it blank if this was supported.

WebBreacher: Understood and the string `html` is too likely to result in false positives.

WebBreacher: For instance, if you submit a username of `yurtufuytf` and visit that page using the `yurtufuytf.wordpress.com` URL, the term `html` appears even though there is no valid user of this name on the platform.

Unpublished: It's still detected as missing because it contains `Do you want to register`. If you still think false positives are too likely it's probably better to skip this PR for now.

WebBreacher: A couple of things to note:

First, not all of the scripts that use WMN may use that account_missing_string to override the positive hit on the account_existence_string (so they might check, is the account there? if so, don't check if it is not there). That is why we use unique strings for the account_existence_string that do not appear on the missing page.

Also, before merging any sites into WMN, I run 2 of the Python-based checkers (`check_online_presence.py` and `web_accounts_list_checker.py`) and see if the content comes back with errors. If it does, then it needs to be reworked before inclusion/merging.

I ran the following and received errors on the `account_missing_code`:

```bash
$ python ./check_online_presence.py -s "WordPress Blog"
 Checking 1 sites
Error when looking up https://Dgwy3LDm3n4oJcHKjaKR.wordpress.com ('account_missing_code')
    As expected, profile found at https://test.wordpress.com
Error when looking up https://unknown31.wordpress.com ('account_missing_code')
```

Unpublished: Thanks for your patience.

I didn't think of varying implementations and only checked with `web_accounts_list_checker.py`.

I'll reopen once I've a more reliable check ready. Most likely after #402 is supported.

PR #408, opened 2022-07-08
Added Friendweb, TotalWar forums, Mmorpg forums

K2SOsint:


web_accounts_list.json
```diff
@@ -4525,6 +4525,39 @@
            "known_accounts" : ["spicerex", "rod-it"],
            "category" : "tech",
            "valid" : true
+          },
+          {
+           "name" : "Friendweb",
```
WebBreacher, line 4530: This site is only available if you are within a certain location (Netherlands). Otherwise it gives a 403 Forbidden error from CloudFlare. I checked with a VPN into NL and I could reach the site. Outside of NL, it does not appear to work. Going to leave it in. Just noting.

WebBreacher:
@K2SOsint please make sure to add yourself to the top of the JSON file as a contributor and to the README.md bottom. Thanks!!

web_accounts_list.json
K2SOsint: I will write this down on my 'to check' list for further additions.

PR #427, opened 2022-07-16
Added 2 game sites

K2SOsint:
forums.thesims.com
discussions.ubisoft.com

Tested & working.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -4711,6 +4711,28 @@
            "known_accounts" : ["amiw", "turbolove"],
            "category" : "social",
            "valid" : true
+          },
+          {
+           "name" : "thesims.com",
+           "check_uri" : "https://forums.thesims.com/en_US/profile/{account}",
+           "account_existence_code" : "200",
+           "account_existence_string" : "The Sims Forums",
+           "account_missing_string" : "User Not Found",
+           "account_missing_code" : "404",
+           "known_accounts" : ["babykittyjade", "simgurumorgan"],
+           "category" : "gaming",
+           "valid" : true
+          },
+          {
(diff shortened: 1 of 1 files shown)


web_accounts_list.json
```diff
@@ -4711,6 +4711,28 @@
...
+           "account_missing_code" : "404",
+           "known_accounts" : ["babykittyjade", "simgurumorgan"],
+           "category" : "gaming",
+           "valid" : true
+          },
+          {
+           "name" : "ubisoft.com",
+           "check_uri" : "https://discussions.ubisoft.com/user/{account}",
+           "account_existence_code" : "200",
+           "account_existence_string" : "Ubisoft Discussion Forums",
```
WebBreacher, line 4730: This is found on the pages with no accounts. Consider changing it to `account-list-navigation`

web_accounts_list.json
```diff
@@ -4711,6 +4711,28 @@
...
+           "account_missing_string" : "Not Found",
```
WebBreacher, line 4731: Let's be more precise here as you could have profiles with the words `Not Found` in them. Consider changing to `<title>Not Found | Ubisoft`

WebBreacher (changes requested):
Please make these changes and rebase you JSON file as it is out of sync with the repo due to other submissions.

web_accounts_list.json
K2SOsint: Will do.

web_accounts_list.json
K2SOsint: Found an <H2 class that is not on the 'Not found' page, will use that as existence string instead.

PR #438, opened 2022-07-18
Update web_accounts_list.json (Adding 1 website)

C3n7ral051nt4g3ncy:
**Adding 1 Website**:

- Pokemon Vortex: https://www.pokemon-vortex.com

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -4779,6 +4779,17 @@
            "valid" : true
           },
           {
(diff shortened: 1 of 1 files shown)


K2SOsint:
I'm getting a 403 error instead of 200 when checking the code. I see that in the JSON file, a lot of Twitter related sites are actually searched using archive.org. Maybe try that?

PR #500, opened 2022-08-21
Update wmn-data.json

C3n7ral051nt4g3ncy:
`m_code` should be below `m_string`


C3n7ral051nt4g3ncy:
Closed due to having submitted the same already.

PR #511, opened 2022-08-26
Added Gigapan site & 'fixed' Picsart

AccentuSoft:
Gigapan site added.

Picsart post_body parameter had a space in it that was removed.

### DIFF
--- a/web_accounts_list.json
+++ b/web_accounts_list.json
@@ -1751,6 +1751,17 @@
       "valid" : true
      },
      {
+      "name" : "Gigapan",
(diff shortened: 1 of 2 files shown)


WebBreacher:
Thanks for this submission. We are still maintaining both the `wmn-data.json` and the `web_accounts_list.json` files for a little bit longer. Can you please add your work to the old JSON file using the old format too?

AccentuSoft:
Done. Some additional lines were changed because of Windows line breaks.

Let me know if you mind so we work around that.

PR #525, opened 2022-08-31
Adding 1 Website

C3n7ral051nt4g3ncy:
## Added `1 Site:`

- Soloby 
- Russian site for learning
- User Example: `http://www.soloby.ru/user/dana`


wmn-data.json
```diff
@@ -4428,6 +4428,18 @@
         "cat" : "social",
         "valid" : true
        },
+       {
+        "name" : "soloby",
+        "uri_check" : "http://www.soloby.ru/user/{account}",
+        "e_code" : 200,
+        "post_body" : "",
+        "e_string" : "- Универ soloBY",
```
WebBreacher, line 4436: The e_string and m_string are very similar. Are they distinct enough?

WebBreacher (changes requested):

wmn-data.json
C3n7ral051nt4g3ncy: @WebBreacher That is all I found, there really wasn't much info after the cURL.
- We can be certain that when a profile is found, the string `- Универ soloBY` is always present
- We can also be sure that when a profile is not found, the `- is missing` 
- I agree that they are not very distinct.

PR #526, opened 2022-08-31
Added Blackbird as a tool that uses WhatsMyName.

p1ngul1n0:


C3n7ral051nt4g3ncy:
> 

@p1ngul1n0 I have not tried Blackbird yet, but it certainly looks good.

PR #575, opened 2022-09-29
Fix twitter wayback, parler wayback, and gettr searches

jocejocejoe:
Detection strings changed for all these site definitions, so fix them up.


WebBreacher:
Thanks for the submission. Can you explain more about what fixes/changes you made to the wayback entries? The space that you removed from each was outside the parameter and was only cosmetic (and was in line with our style)

jocejocejoe:
The changes I made are to the _values_ for e_string and m_string, which happen to be looking at JSON returned from the wayback machine.  It is not changing the whitespace in wmn-data.json itself.

i.e. wayback was returning a space before the colon, and now it is not. It went from:

`"archived_snapshots" : {"closest",`
to
`"archived_snapshots": {"closest",`

so this needs to be fixed up for detection to work properly.  (we can't use regexps in the detection strings, right?)

WebBreacher:
Apologies for the confusion. I was looking at the changes on my mobile and the strings were wrapping at just the wrong point.

jocejocejoe:
Now my turn to apologize, it seems Gettr isn't working. I swear I saw claimable true/false working before I submitted it. Now it's returning random usernames as present, and returning a http_status_code of -1? I'm confused, but looking into it. Sorry!

PR #600, opened 2022-10-18
Added the BabyPips forum

swedishmike:
Added the forum from the BabyPips site. Marked it as 'social' since it's a forum, potentially possible to be 'finance' too?


wmn-data.json
```diff
@@ -518,6 +518,18 @@
         "cat" : "XXXPORNXXX",
         "valid" : true
        },
+       {
+        "name" : "BabyPips",
+        "uri_check" : "https://forums.babypips.com/u/{account}/summary",
```
WebBreacher, line 523: Can you shift this to `uri_pretty` and make the `https://forums.babypips.com/u/baemax023.json` the `uri_check`? It'll be MUCH easier to grab consistent data from the JSON and then point users at the HTML.

PR #611, opened 2022-10-31
Adding 1 site

C3n7ral051nt4g3ncy:
## Added `1 site`: 

- `counter_social`

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -3648,7 +3648,7 @@
         "uri_check" : "https://{account}.myportfolio.com/work",
         "post_body" : "",
         "e_code" : 200,
-        "e_string" : "digital artist, illustrator, game art",
+        "e_string" : "class="page-title",
         "m_string" : "<title>Adobe Portfolio | Build your own personalized website</title>",
         "m_code" : 302,
         "known" : ["artkonina", "fox"],
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -3648,7 +3648,7 @@
         "uri_check" : "https://{account}.myportfolio.com/work",
         "post_body" : "",
         "e_code" : 200,
-        "e_string" : "class="page-title",
+        "e_string" : "class=\"page-title",
         "m_string" : "<title>Adobe Portfolio | Build your own personalized website</title>",
         "m_code" : 302,
         "known" : ["artkonina", "fox"],
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -4029,6 +4029,18 @@
         "known" : ["radio357", "radionowyswiat"],
         "cat" : "finance",
         "valid" : true
+        },        
+        {
+        "name" : "Paypal",
+        "uri_check" : "https://www.paypal.com/paypalme/{account}",
+        "post_body" : "",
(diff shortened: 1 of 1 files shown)


C3n7ral051nt4g3ncy:
## Added `2nd site`:

- `sofurry`

wmn-data.json
```diff
@@ -1240,6 +1240,18 @@
         "cat" : "art",
         "valid" : true
        },        
+       {
+        "name" : "counter_social",
+        "uri_check" : "https://counter.social/@{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "@counter.social) - CounterSocial</title>",
+        "m_string" : "The page you are looking for isn't here.",
```
WebBreacher, line 1249: Can you please verify the `m_code`? In Chrome I see the `'` is `&39;` so the `m_code` value should be `The page you are looking for isn&#39;t here.`

C3n7ral051nt4g3ncy:
## Added `3rd site`:

- `myportfolio`

C3n7ral051nt4g3ncy:
## Added `4th site`: 

- `Paypal`

wmn-data.json
```diff
@@ -3629,6 +3641,18 @@
...
         "known" : ["Djelizamay","Andysparkles"],
         "cat" : "social",
+        "valid" : true
+       },        
+       {
+        "name" : "myportfolio",
+        "uri_check" : "https://{account}.myportfolio.com",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "html lang",
```
WebBreacher, line 3651: Please choose a more distinct string here. Some of our targets and users may not be english readers and this value could be dynamic. Perhaps something like `preserve-whitespace">Back to Top` and add `/work` to the end of the `uri_check`?

wmn-data.json
```diff
@@ -4004,6 +4028,18 @@
...
         "cat" : "finance",
+        "valid" : true
+        },        
+        {
+        "name" : "Paypal",
+        "uri_check" : "https://www.paypal.com/paypalme/{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "userInfo",
+        "m_string" : "Get your very own PayPal.Me link",
```
WebBreacher, line 4039: Let's try to make the `m_string` a component of the page not text shown in the browser as people that use languages other than English will get false positives.

wmn-data.json
```diff
@@ -3629,6 +3641,18 @@
...
         "known" : ["Djelizamay","Andysparkles"],
         "cat" : "social",
+        "valid" : true
+       },        
+       {
+        "name" : "myportfolio",
+        "uri_check" : "https://{account}.myportfolio.com/work",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "digital artist, illustrator, game art",
```
WebBreacher, line 3651: This is something that varies based upon the user account. For instance, https://bqwphotography.myportfolio.com does not have those data points.

WebBreacher: How about `class="page-title"`

PR #613, opened 2022-11-01
Adding 4 sites 

C3n7ral051nt4g3ncy:
## Added `4 sites`:

- `resumes_actorsaccess`
- `mymfans`
- `likeevideo`
- `gfycat`

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -2270,13 +2270,14 @@
         },        
         {
         "name" : "gfycat",
-        "check_uri" : "https://gfycat.com/@{account}",
-        "account_existence_code" : "200",
-        "account_existence_string" : "(@",
-        "account_missing_string" : "<h1>Page not found</h1>",
-        "account_missing_code" : "404",
-        "known_accounts" : ["timviechanoi", "sannahparker"],
-        "category" : "misc",
+        "uri_check" : "https://gfycat.com/@{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "(@",
+        "m_string" : "<h1>Page not found</h1>",
+        "m_code" : 404,
+        "known" : ["timviechanoi", "sannahparker"],
+        "cat" : "misc",
         "valid" : true
        },
        {
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -3121,7 +3121,7 @@
         "m_string" : "Unable to find user",
         "m_code" : 404,
         "known" : ["owlthorns", "Severkosh"],
-        "category" : "social",
+        "cat" : "social",
         "valid" : true
        },
        {
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -3637,7 +3637,7 @@
         "post_body" : "",
         "e_code" : 200,
         "e_string" : " • MYM</title>",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -2266,6 +2266,18 @@
...
         "known" : ["gettr", "support"],
         "cat" : "social",
+        "valid" : true
+       },        
+       {
+        "name" : "gfycat",
+        "uri_check" : "https://gfycat.com/@{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "(@",
```
WebBreacher, line 2276: This `e_string` is too simple and runs the risk of having false positives. Suggest using `gfycat-username`

wmn-data.json
```diff
@@ -3098,6 +3110,18 @@
...
         "known" : ["mohammed01", "mohammed03"],
         "cat" : "gaming",
+        "valid" : true
+       },        
+       {
+        "name" : "likeevideo",
+        "uri_check" : "https://likee.video/@{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "(@",
```
WebBreacher, line 3120: String is too simple and likely to cause false positives. Suggest `https://img.like.video/`. It looks like the missing profile pages use the HTTP and the pages with profiles use HTTPS for this URL. Please check.

wmn-data.json
```diff
@@ -3098,6 +3110,18 @@
...
+        "m_string" : "Unable to find user",
```
WebBreacher, line 3121: Looks like this `Unable to find user` is not in the returned HTML but in a JSON file used for errors. Please recheck this and find another `m_string` that is in the HTML.

wmn-data.json
```diff
@@ -3605,6 +3629,18 @@
...
         "cat" : "health",
+        "valid" : true
+       },        
+       {
+        "name" : "mymfans",
+        "uri_check" : "https://mym.fans/{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : " • MYM</title>",
+        "m_string" : "Il n’y a rien ici…",
```
WebBreacher, line 3640: This string is specific to you since you are probably viewing the site in a browser/commandline with French language/from French IP address space. In my browser, I see the error `There's nothing here...`. Perhaps use the title HTML tag `• MYM • Exclusive social network for creators &amp; fans</title>`

C3n7ral051nt4g3ncy: @WebBreacher Thanks for your help, `Done!` re change of `m_string` on `mymfans`

wmn-data.json
C3n7ral051nt4g3ncy: @WebBreacher: `Done` re change of `e_string` for `gfycat`

wmn-data.json
C3n7ral051nt4g3ncy: @WebBreacher : `Done!` re change of `e_string` for `likeevideo`

wmn-data.json
C3n7ral051nt4g3ncy: @WebBreacher: `Done!` re change of `m_string` for `likeevideo`

Issue #583, opened 2022-10-05
How do people know what the "platform" names are for useragents?


WebBreacher:
I see in the `whatsmyname/app/resources/user-agents.json` file there is JSON with the name and value pairs for different platforms/useragents. How do people know what those values are by running the script?

Also not sure how, once I know that I want "Chrome on macOS" to pass that to the script.

cc: @yooper

yooper:
That is a great question, it is sort of an undocumented feature at the moment.

yooper:
Running the command,
```
 python whats_my_name.py -h
```
does list the feature. 

The default, user agent is _"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"_

I will make a small PR that updates the documentation returned when running the program with the `-h` flag. The documentation will specify what the default user agent platform is.

WebBreacher:
I'm more talking about the multipl UAs. It'd be nice to have a `1` = `Windows Firefox`, `2` = `macOS Chrome` etc in the help to match what arguments the users can pass to the script. `-ua 1` etc.

WebBreacher:
Resolved.

PR #628, opened 2022-11-10
Fixes issue #583

yooper:
You can now set the user agent using an int.
`python whats_my_name.py -ua 2`

To see what the ids are, check out the updates to this file.
`whatsmyname/app/resources/user-agents.json`


WebBreacher:
Thank you @yooper

PR #636, opened 2022-11-16
Adds spotify back in.

ppanthony:
From what I can tell looks like this is no longer false positiving. If there's any further detail on accounts which are exhibiting false positive behaviour happy to look at this.

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5131,6 +5131,18 @@
         "valid" : true
        },
        {
+        "name" : "spotify",
+        "uri_check" : "https://open.spotify.com/user/{account}",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "on Spotify",
+        "m_string" : "",
+        "m_code" : 404,
+        "known" : ["spotify"],
+        "cat" : "social",
+        "valid" : true
+       },
+       {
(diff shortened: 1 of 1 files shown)


C3n7ral051nt4g3ncy:
@ppanthony Good one! ! 🚀

C3n7ral051nt4g3ncy (changes requested):
@ppanthony : There should be at least 2 known accounts.

C3n7ral051nt4g3ncy:
@ppanthony: Noticed Spotify is already in the` wmn-data.json`

WebBreacher:
Yes. Thank you @C3n7ral051nt4g3ncy for noticing. We have an automated checker that should have caught this too. I will research that. @ppanthony it looks like since we are already checking spotify and the URL is the same as your's, this PR may not be necessary. Are you finding the existing Spotify entry not sufficient?

ppanthony:
@WebBreacher @C3n7ral051nt4g3ncy - my bad. Didn't notice that it was in there, was just looking to contribute to the project and noticed the previous 'removed' discussion here: https://github.com/WebBreacher/WhatsMyName/wiki/Problem-Removed-Sites of which spotify was listed. You can close the pull! Sorry again!

Issue #560, opened 2022-09-16
HTTP POST entry and output


WebBreacher:
Related to #555 

POST - [https://forumserver.twoplustwo.com](https://forumserver.twoplustwo.com/ajax.php?do=usersearch) site has been added but:

There is no real negative result. The positive result has extra data but the negative result has the same HTTP response code and same data except is missing the positive string (`e_string`)

C3n7ral051nt4g3ncy:
ok

C3n7ral051nt4g3ncy:
@WebBreacher : Is there anything else to do regarding this issue or can it be closed?

WebBreacher:
Yes it should stay open until the HTTP POST checking output is resolved by @yooper

yooper:
We haven't implemented the POST functionality, yet in WMN. We do need to verify and confirm that the example, found in #555, is how we want POST data, within the `wmn.json` file.

WebBreacher:
I believe that format for the data ` securitytoken=guest&do=usersearch&fragment=Redsoxn` would be preferred as that should be able to be cut and pasted into the data portion of the POST.

yooper:
I plan on starting on this functionality this week.

PR #647, opened 2022-12-10
adds ability to post to sites to test for a user, closes #558, #560

### LINKED ISSUE #558: HTTP POST Site Output
image
@yooper - I'd like to work with you on more meaningful output for the HTTP POST case. We cannot just echo the URL since the data is not appearing in the URL. Also, I'm not sure POSTs are being eval'd correctly.

yooper:
@WebBreacher , while adding this feature, I could not get a conclusive answer if the user existed. We need more sample POST cases in our dataset to verify this feature works correctly. @C3n7ral051nt4g3ncy , you have any test/sample POST sites I can develop against? 

To run the example
```
whats_my_name.py -u redsox -s twoplustwo --verbose -a
```
Here is the curl command I used to test the functionality, but did not see any mention of the user 'redsox' in the output. 

```
 curl -o -d "securitytoken=guest&do=usersearch&fragment=adam" -X POST https://forumserver.twoplustwo.com/search.php > /tmp/test
 grep redsox /tmp/test
```

Cheers,
--Dan

### DIFF
--- a/whatsmyname/app/tasks/process.py
+++ b/whatsmyname/app/tasks/process.py
@@ -159,20 +159,41 @@ async def request_worker(session: ClientSession, cli_options: CliOptionsSchema, 
     headers = {
         'User-Agent': site.user_agent
     }
+    if site.post_body:
+        try:
+            post_body_altered: str = site.post_body.replace('{account}', site.username)
+            form_data = {x[0] : x[1] for x in [x.split("=") for x in post_body_altered[1:].split("&") ]}
+            async with session.post(site.generated_uri,
+                                   data=form_data,
+                                   timeout=cli_options.per_request_timeout,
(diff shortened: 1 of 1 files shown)


WebBreacher:
Check out the POST for "SpankPay" as that uses POST and API.

yooper:
The example given doesn't have any data in the `post_body` payload. 
```
       {
        "name" : "SpankPay",
        "uri_check" : "https://api.spankpay.com/graphql",
        "uri_pretty" : "https://spankpay.me/{account}",
        "post_body" : "NEED TO FIX THIS",
        "e_code" : 200,
        "e_string" : "headerPhoto",
        "m_string" : "User not found",
        "m_code" : 404,
        "known" : ["Medroxy","brittanyandrews"],
        "cat" : "finance",
        "valid" : false
       }
```

PR #663, opened 2023-01-19
Fix VK detection

ar-2001:
Fix issue with vk.com detection. Now it responds with the title in Russian by default. 

(also fixed 2 lines with ^M windows endlines)


WebBreacher:
I think we need a different `e_string` as the title still appears in English for my American web browser. I'll bet it depends on where your system comes from or system settings.

Could you find a different `e_string` in the page that does not change?

WebBreacher:
Works for me. Thanks for making that change and for submitting this PR! Really appreciate it!

PR #666, opened 2023-01-20
Fix TikTok detection. Use oembed API instead of main site

ar-2001:
I had problems with tiktok detection. Sometimes there were timeouts (it depends on User Agent), in other times there were almost empty page with javascript. Looks like everything is dynamically loaded now.

After some research I found open API, that doesn't require tokens. It's used for embedding profiles. as I understood. Not sure how it's stable, and how long it will work, but it works well for detection.

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -6405,6 +6405,19 @@
         "valid" : true
        },
        {
+        "name" : "TikTok",
+        "uri_check" : "https://www.tiktok.com/oembed?url=https://www.tiktok.com/@{account}",
+        "uri_pretty" : "https://www.tiktok.com/@{account}?lang=en",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "author_url",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -6404,6 +6404,19 @@
...
         "valid" : true
        },
+       {
+        "name" : "TikTok",
+        "uri_check" : "https://www.tiktok.com/oembed?url=https://www.tiktok.com/@{account}",
+        "uri_pretty" : "https://www.tiktok.com/@{account}?lang=en",
+        "post_body" : "",
+        "e_code" : 200,
+        "e_string" : "author_url",
+        "m_string" : "Couldn't find this account",
```
WebBreacher, line 6414: I'm getting `"message": "Something went wrong",` when I do an account that is not found. Can you recheck this? Also getting `"code": 400`.

WebBreacher:
Thanks for finding this gem!!!

PR #672, opened 2023-02-04
Change eToro uri

C3n7ral051nt4g3ncy:


wmn-data.json
```diff
@@ -1970,7 +1970,7 @@
        },
        {
         "name" : "eToro",
-        "uri_check" : "https://www.etoro.com/people/{account}",
+        "uri_check" : "https://www.etoro.com/api/logininfo/v1.1/users/{account}",
```
WebBreacher, line 1973: Is there a `pretty_uri` you can add here that has a direct link to the user's profile?

C3n7ral051nt4g3ncy: yep

C3n7ral051nt4g3ncy: @WebBreacher : done ☑️

PR #680, opened 2023-02-12
Add Truth Social and patriots.win

jocejocejoe:
Truth Social now seems to allow unauthenticated lookups


WebBreacher:
Looks llike it. Thanks for adding this back!

PR #681, opened 2023-02-12
Fix anonup detection

jocejocejoe:
Following</span> no longer works


WebBreacher:
Thanks for fixing this.

PR #688, opened 2023-03-06
Invalid name in username

yooper:
An optional attribute string, `invalid_chars` can be provided within the `wmn-data.json` file. This string is broken into characters and each character is removed from the username for that given site. A unit test case was written and you can test the functionality with the following command.

```bash
python whats_my_name.py -u john.doe -s Wanelo -a
```


WebBreacher:
Before I merge this, I'll need to update the JSON Schema files or the GitHub actions that check for valid JSON files will break. Thanks for this submission....lemme do that work tonight.

WebBreacher:
@yooper hate to do this to you but can you rebase your PR? I just made some schema changes and other changes to the JSON and your PR has confliuts.

yooper:
@WebBreacher , I updated the PR with your changes from main. It should be fine to merge.

WebBreacher:
Thanks @yooper!

PR #694, opened 2023-03-17
Adding 1 site 

C3n7ral051nt4g3ncy:
**Added site**: 

- `GoodGame` (Russian Gaming site)


wmn-data.json
```diff
@@ -2459,6 +2459,18 @@
         "m_code" : 400,
         "known" : ["johnny", "dev"],
         "cat" : "coding",
+        "valid" : true
+       },
+       {
+        "name" : "Goodgame_Russia",
+        "uri_check" : "hhttps://goodgame.ru/channel/{account}/",
+        "post_body" : "",
```
WebBreacher, line 2467: Please remove this line. We changed this to an optional parameter that should only be added when there is POST data.

C3n7ral051nt4g3ncy: @WebBreacher ✔️

PR #700, opened 2023-05-15
Fix implicit encoding that breaks cross platform usage

janldk:
The current code does not specify the encoding when opening files to read/write. Besides breaking one of the Zen of Python-rules (_explicit is better than implicit_) this breaks cross platform as the encoding then is platform dependent (whatever `locale.getencoding()` returns). If this is `cp1252` - which is often the case on Windows - the script crashes. After fixing it on the loading of `wmn-data.json` (https://github.com/WebBreacher/WhatsMyName/blob/7e6ac66d0515e091782f02b3488f1fdbfc316ae9/whatsmyname/app/extractors/file_extractors.py#L16) is actually crashes silently, leading the user to believe that the script does not return any hits. This pull request should fix the issue in every needed part of the library.

### DIFF
--- a/whatsmyname/app/extractors/file_extractors.py
+++ b/whatsmyname/app/extractors/file_extractors.py
@@ -13,7 +13,7 @@ def site_file_extractor(file_path: str) -> List[SiteSchema]:
     :param file_path: str
     :return: List[Site]
     """
-    with open(file_path) as json_file:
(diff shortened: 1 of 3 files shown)


WebBreacher:
Closing this as the script is no longer part of this project.  Yooper's script is now at https://github.com/yooper/reveal-my-name

PR #705, opened 2023-05-31
Update wmn-data.json

janldk:
Fix broken site


wmn-data.json
```diff
@@ -159,7 +159,7 @@
         "name" : "adultism",
         "uri_check" : "https://www.adultism.com/profile/{account}",
         "e_code" : 200,
-        "e_string" : "static/r-1OqQ4o/css/www/main.css",
+        "e_string" : "profile-info",
```
WebBreacher, line 162: Can you change this to something that only appears once on the page with the account? Something like `last login:` or `Member since:`

wmn-data.json
```diff
@@ -193,7 +193,7 @@
         "name" : "Airline_Pilot_Life",
         "uri_check" : "https://airlinepilot.life/u/{account}",
         "e_code" : 200,
-        "e_string" : "<title> Profile -",
+        "e_string" : "<title>  Profile -",
```
WebBreacher, line 196: for this site, we can use the `https://airlinepilot.life/u/{account}.json` for the `uri_check` and the `https://airlinepilot.life/u/{account}` for the `uri_pretty`. Then can pick one of the common parameters from the JSON like `primary_group_name` or `trust_level`. If you don't want to do this rework, I'm happy to.

wmn-data.json
janbinx: Sure. I changed it with commit 4daf7aba7d537a64003bfa54800fd152e993400c

wmn-data.json
janbinx: That'll work too. As will this fix.

PR #710, opened 2023-06-29
litmind.club does not seem to exist any longer?

swedishmike:
I cannot resolve litmind.club - tried on different resolvers so it's not my ad blocker.

For now I've marked it as not valid, if you cannot resolve it either I can of course remove it completely and re-PR.


WebBreacher:
I too cannot resolve it. If you wouldn't mind removing the whole site and resubmitting the PR, I would appreciate it.

PR #718, opened 2023-07-06
Fixed existing sites

swedishmike:
Here is a first batch of fixed and deleted sites. I've tested that they work with the existing known usernames, and have triggered 'not found' by testing with obviously bad names.


WebBreacher:
Thank you for these submissions.

PR #719, opened 2023-07-08
AllTrails - removed, 403's now. Seems to have locked viewing members …

swedishmike:
…without being logged in.


WebBreacher:
Thank you for this work. Please put the sites you remove into https://github.com/WebBreacher/WhatsMyName/wiki/Problem-Removed-Sites

PR #721, opened 2023-07-09
A bit of a grab bag of changes, majority 'just' updates of known user names.

swedishmike:
There are a few updated known usernames here, tested and verified so should be ok.

Then there is the following:

- `Cafecito` - didn't work, verified by testing via whatsmyname.app too, there was what looked like a control character in the code when viewed in an editor. Removed that and replaced with a 'clean' space and it worked again.
- `Clubhouse` - didn't work, verified by testing via whatsmyname.app too. Had to replace the e_string and now it works again.
- `Mastodon-counter.social` - as discussed yesterday; behind CloudFlare and redirect now so removed. Verified via whatsmyname.app that it doesn't work there any longer too.
- `CloudFlare` - 403's, might even be behind 'itself' so to speak. Removed. Verified via whatsmyname.app that it doesn't work there any longer too.

If/when you ok the removals and merge I'll add those two to the 'removed sites wiki' too.

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -1131,7 +1131,7 @@
         "e_string" : "<title>  Profile - ",
         "m_string" : "Oops! That page doesn’t exist",
         "m_code" : 404,
-        "known" : ["doctypeme", "jon_morris"],
+        "known" : ["doctypeme", "ocean.war"],
         "cat" : "coding",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Thank you!

PR #724, opened 2023-07-13
update the m_string for GIthub, Gitlab and Foursquare

janldk:
The current `m_string` for missing accounts are wrong. I've updated the ones I get, when using HTTP clients (`curl` & `requests`).


wmn-data.json
```diff
@@ -2311,7 +2311,7 @@
         "uri_check" : "https://gitlab.com/{account}",
         "e_code" : 200,
         "e_string" : "Member since",
-        "m_string" : "GitLab.com offers free unlimited",
+        "m_string" : "You are being <a href="https://gitlab.com/users/sign_in">redirected</a>",
```
WebBreacher, line 2314: Looks like you have an error in the JSON here with an unescaped double quote `"`. Can you change the `m_string` to not have a double quote in it?

janbinx: ah, how stupid :) Fixed with cfdc055db0f55d8c6f34ef27a0545d63523132bf

wmn-data.json
```diff
@@ -2066,7 +2066,7 @@
         "uri_check" : "https://foursquare.com/{account}",
         "e_code" : 200,
         "e_string" : "on Foursquare</title>",
-        "m_string" : "Foursquare - Independent Location Data Platform",
+        "m_string" : "302 Found</title>",
```
WebBreacher, line 2069: I'm not seeing this in the response. Can you confirm and explain how you are checking this?

liamb13: For me, it's redirecting to the home page. With the page title: "Location Technology Unlocking Powerful Connections | FSQ"

liamb13: Tested with node (axios) and python (requests) and received the same. Redirects to home with: <title>Unlocking Powerful Consumer Connections with Location | FSQ</title>

wmn-data.json
```diff
@@ -2300,7 +2300,7 @@
         "uri_check" : "https://github.com/{account}",
         "e_code" : 200,
         "e_string" : "p-nickname vcard-username d-block",
-        "m_string" : "<title>Page not found ",
```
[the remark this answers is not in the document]
liamb13: Python isn't my language of choice, but trying:
```
import requests

url = "http://github.com/abcdeffsafas"

payload = {}
headers = {
    'Accept-Encoding': 'identity',
}

response = requests.get(url, headers=headers, data=payload)

print(response.text)

```
Returns 'Not Found'

```
print(response.status)
```

Returns 404.

Not sure if this is at all helpful.

WebBreacher:
Sorry to do this but I'm traveling and cannot check more into these submissions. The PR will need to remain open for a little bit until I return.

wmn-data.json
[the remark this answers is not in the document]
WebBreacher: Try `curl 'https://github.com/abcdeffsafas'   -H 'authority: github.com'   -H 'accept: text/html'` and you get the entire HTML. Without that "text/html" header part, it'll only give the Not Found. Problem is that, if our users are using tools like whatsmyname.app, they will be sending the header data and will get a different result.

This is not that big a deal in reality, since most tools do not use the `m_string` at all.

In the web browser, the response contains `not found` but not `Not Found`. On the command line, it is reversed. So if  a tool does case sensitive matching of the responses, it will get false positives/negatives depending on how the site is checked. 

Since most of the users of this project use the web browser version AND since most tools don't even check the `m_string` I'd like to leave the string as it is.

PR #744, opened 2023-08-02
Update wmn-data.json

liamb13:
https://projecthub.arduino.cc returns 200 for accounts not found.

Updated it's m_code and m_string


WebBreacher:
Thanks for the fix!

PR #745, opened 2023-08-17
feat: replit.com

msramalho:
Adds support for replit.com.

---
Also tried notion.so but that would only work with a Selenium or similar browser approach. 

This repo used to have code for testing the new platforms and not any more, I went into the /dev branch and fug out the old `check_online_presence.py` and had to replace the renamed fields but that was useful for testing with `python check_online_presence.py -in wmn-data.json -s Replit`


wmn-data.json
```diff
@@ -4987,6 +4987,17 @@
         "cat" : "XXXPORNXXX",
         "valid" : true
        },
+       {
+        "name" : "Replit",
+        "uri_check" : "https://replit.com/@{account}",
+        "e_code" : 200,
+        "e_string" : "Copy profile link",
+        "m_code" : 404,
+        "m_string" : "404 Not Found",
```
WebBreacher, line 4996: I'm not seeing this string in the response. Perhaps use `<title>Replit - 404 - Replit</title>` instead? Although, that only is returned via a web browser request. A `curl` request for an incorrect user has a response of `{"message":"User not found","name":"NotFoundError","status":404}`

WebBreacher:
Re: the removal of the python script....yes, we removed all code from this project and are focusing on just the JSON file. There are "checker" scripts on the README.md page that have scripts for checking WMN data.

wmn-data.json
msramalho: Thanks for raising it, I trusted the check script to strongly. 
So I had the same response from CURL vs browser, I'll modify it to the browser version.

msramalho:
Re: python scripts, so you'd avise just using browser version vs curl responses?

WebBreacher:
Know what? Let's do both. Make 2 entries for this 1 site.

Change this to 1 entry from `replit cli` and 1 `replit web`. That way each set of users can benefit.

Issue #760, opened 2023-10-12
change chess.com to use api


WebBreacher:
https://api.chess.com/pub/player/hikaru

WebBreacher:
Closed with #769

PR #769, opened 2023-11-03
change chess.com to use api

grabowskiadrian:
#760

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -974,10 +974,10 @@
        },
        {
         "name" : "Chess.com",
-        "uri_check" : "https://www.chess.com/member/{account}",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -974,10 +974,10 @@
        },
        {
         "name" : "Chess.com",
-        "uri_check" : "https://www.chess.com/member/{account}",
+        "uri_check" : "https://api.chess.com/pub/player/{account}",
         "e_code" : 200,
```
WebBreacher, line 978: Thanks for switching this. When we use an API for the check, we always try to add a `uri+pretty` that an app can point a user to. Can you just add the previous `uri_check` as a `uri_pretty`?

PR #811, opened 2024-05-09
Internet Archive User Search fix

balestek:
Using Internet Archive advanced search for the json output to fix the 302 redirection with the normal search
e_string and m_string updated
uri_pretty added


WebBreacher:
thanks for the PR!

balestek:
No problem. I'll continue if I can fix / add sites.

PR #828, opened 2024-06-13
Discogs, Virustotal and Wetransfer fix

balestek:
**VirusTotal**

- Uses an undocumented endpoint to circumvent the profile page javascript loader.
- Headers are required, values can be dummy data, even for "X-VT-Anti-Abuse-Header" (works as well with "1"). But near expected values is more stealth.
- The I (maj "i") in the key "Accept-Ianguage" is on purpose, planted by VirusTotal.

**Discogs**

- Uses discogs official api to circumvent the page javascript loader.
- Official api rate limit is 25 requests per minute for unregistered user (by IP). It should be ok.
- Changed known accounts because they were not working (usernames changed by user)

**Wetransfer**

- e_string and m_code changed

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5942,11 +5942,18 @@
       },
       {
         "name" : "Virustotal",
-        "uri_check" : "https://www.virustotal.com/gui/user/{account}",
+        "uri_check" : "https://www.virustotal.com/ui/users/{account}",
+        "pretty_uri" : "https://www.virustotal.com/gui/user/{account}",
+        "headers" : {
+                "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:127.0) Gecko/20100101 Firefox/127.0",
+                "X-Tool": "vt-ui-main",
+                "Accept-Ianguage": "en-US,en;q=0.9,es;q=0.8",
+                "X-VT-Anti-Abuse-Header": "MTAxOTFwMDcxOTEtWkc5dWRDQmlaU0JsZG2scy5xNzE4Mjc1NDI0LjUzMw=="
+        },
         "e_code" : 200,
-        "e_string" :"USER PROFILE",
-        "m_string" : "User not found",
-        "m_code" : 200,
+        "e_string" :"data",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -5942,11 +5942,18 @@
       },
       {
         "name" : "Virustotal",
-        "uri_check" : "https://www.virustotal.com/gui/user/{account}",
+        "uri_check" : "https://www.virustotal.com/ui/users/{account}",
+        "pretty_uri" : "https://www.virustotal.com/gui/user/{account}",
+        "headers" : {
+                "User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:127.0) Gecko/20100101 Firefox/127.0",
```
WebBreacher, line 5948: In checking this detection I noticed the ui/users uri_pretty requires the solving of a captcha. I'm not sure that using this detection will work since the one in the `headers` param is expired. Thoughts?

```
  "error": {
    "code": "RecaptchaRequiredError",
    "message": "Please re-send request with a valid reCAPTCHA response in the \"x-recaptcha-response\" header"
```

balestek:
Ok, I stop with theses 3

wmn-data.json
balestek: It seems that Virustotal only checks if the header is present. It doesn't test the x-racaptcha-response at this point (which is X-VT-Anti-Abuse-Header). It passes with all headers values equals to "1".

balestek: So, I put a string that looks like a token, but it could be anything actually.

WebBreacher: Thanks for the explanation. My concern is that clients like whatsmyname.app don't pass the header statements but that is not our concern but one for the downstream clients.

WebBreacher:
Thanks for the submissions!

wmn-data.json
balestek: Yes, as several checks require headers (and maybe more in the future), whatsmyname.app might need to add headers to the client browser requests.

PR #830, opened 2024-06-25
ok.ru fix

balestek:
m_string fixed

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -3811,7 +3811,7 @@
         "uri_check" : "https://ok.ru/{account}",
         "e_code" : 200,
         "e_string" : "| OK",
-        "m_string" : "This page does not exist on OK",
+        "m_string" : "/error404",
         "m_code" : 404,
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -3811,7 +3811,7 @@
         "uri_check" : "https://ok.ru/{account}",
         "e_code" : 200,
         "e_string" : "| OK",
-        "m_string" : "This page does not exist on OK",
+        "m_string" : "/error404",
```
WebBreacher, line 3814: It looks like this `m_string` is only present in commandline/python-based requests and not from web browsers. Can you confirm this?

balestek: Hello,

"This page does not exist on OK" doesn't exists with a python request. It's in cyrillic "Этой страницы нет в OK".
About web browsers, you think about whatsmyname.app ?
I've just found that "class="p404_t" exists on both. I let you change that ?

balestek: Ok, I've updated it

PR #840, opened 2024-07-31
Add Poe.com

maxk096:


WebBreacher:
Thanks for the submission but I'm very sure this will not work within our project due to the site being being Cloudflare's anti-bot features. Can you test visiting to https://poe.com/profile/spaustin without getting Cloudflare prompted?

maxk096:
I've visited the link you provided and didn't receive Cloudflare prompts in Chrome and Firefox using regular and incognito versions. And I don't recall seeing such prompts before on this website.

WebBreacher:
Unfortunately, I'm getting them which means that Cloudflare is present. We can add this but I do not think it'll work in browsers or Python tooling due to CF.
image

wmn-data.json
```diff
@@ -4280,6 +4280,16 @@
         "known" : ["ftdl", "ducensor"],
         "cat" : "social"
        },
+       {
+        "name" : "Poe.com",
+        "uri_check" : "https://poe.com/profile/{account}",
+        "e_code" : 200,
+        "e_string" : "\"__isNode\": \"PoeUser\"",
```
WebBreacher, line 4287: Can you please remove the space between the `: \"` so this is `\"__isNode\":\"PoeUser\"`

WebBreacher: In both commandline curl and web browser, I see no space.

maxk096: Removed the spaces, and in m_string as well

WebBreacher:
thanks for the submission!

PR #842, opened 2024-08-01
Replaced spaces with underscores and renamed "xx NSFW xx" to "NSFW" in name fields to remove spaces

Itsoon:
Sorry, I formatted the file when I made the changes, if that's a problem, I can recommit without the formatting.


Itsoon:
I renamed "xx NSFW xx" to "NSFW" first to remove the spaces, then I thought NSFW already meant what it said, so why add the "xx". But if it's more visible as you say, why not leave it and remove the spaces. "xxNSFWxx". Sounds like the minecraft trend in 2012 but at least it's visible.
Or we can leave "xx NSFW xx" like that with spaces to force the use of quotation marks when typing the category in the terminal.

I'm going to redo a commit without formatting the json to see the changes.

WebBreacher:
Sorry for not being clear above....I am more curious about what caused you to do this? Were you solving a 'problem'? You hint at having an issue in a command line tool, yes?

Itsoon:
I did this because, in my opinion, the format of the names is neither uniform nor consistent. At one point, the name contains spaces, at another it's capitalized, at another there's .com in the name, and at another there's an underscore. With this commit, I wanted to homogenize the naming.

WebBreacher:
OK. I understand that re: the name as we try to use the name of the site as they show it on their site instead of a modified version. This is a valid point, I believe. Maybe we should move the `name` to something more uniform like `domain`.

Why change the `xx NSFW xx`?

Itsoon:
I replied above, you must have missed the message.

WebBreacher:
Opened issue #843 to address this valid concern. Thanks for suggesting it!

Itsoon:
> Opened issue #843 to address this valid concern. Thanks for suggesting it!
👌

PR #845, opened 2024-08-06
Added Letterboxd

serdaraltin:
Added Letterboxd

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -6560,5 +6560,15 @@
         "known" : ["carlito", "taylor"],
         "cat" : "social"
       }
+      {
+        "name" : "Letterboxd",
+        "uri_check" : "https://letterboxd.com/{account}",
+        "e_code" : 200,
+        "e_string" : "RECENT ACTIVITY",
+        "m_string" : "Sorry, we can’t find the page you’ve requested.",
+        "m_code" : 404,
+        "known" : ["serdaraltin", "jeaba"],
+        "cat" : "social"
+      }
   ]
 }


wmn-data.json
```diff
@@ -6560,5 +6560,15 @@
         "known" : ["carlito", "taylor"],
         "cat" : "social"
       }
+      {
+        "name" : "Letterboxd",
+        "uri_check" : "https://letterboxd.com/{account}",
```
WebBreacher, line 6565: Can you add a `/` to the end of the `uri_check` as I'm getting redirected without it. So it should be `https://letterboxd.com/{account}/`

wmn-data.json
```diff
@@ -6560,5 +6560,15 @@
         "known" : ["carlito", "taylor"],
         "cat" : "social"
       }
+      {
+        "name" : "Letterboxd",
+        "uri_check" : "https://letterboxd.com/{account}",
+        "e_code" : 200,
+        "e_string" : "RECENT ACTIVITY",
```
WebBreacher, line 6567: I tried via web browser and curl and cannot see the `RECENT ACTIVITY` `e_string` in any response. I get a good response but don't see that string. I'd suggest using the string `’s profile on Letterboxd` which appears in the title tag.

wmn-data.json
```diff
@@ -6560,5 +6560,15 @@
         "known" : ["carlito", "taylor"],
         "cat" : "social"
       }
```
WebBreacher, line 6562: You need to change this to be `},` please

WebBreacher:
Please also move this to the "L" section of the JSON file right after `        "name" : "kwejk.pl",` so it is alphabetized

PR #847, opened 2024-08-11
Update wmn-data.json

janhalendk:
https://story.snapchat.com/s/ redirects to https://www.snapchat.com/add/ - also the `m_string` is full upper case

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5009,12 +5009,12 @@
         "cat" : "social"
        },
        {
-        "name" : "Snapchat Stories",
-        "uri_check" : "https://story.snapchat.com/s/{account}",
+        "name" : "Snapchat",
+        "uri_check" : "https://www.snapchat.com/add/{account}",
         "e_code" : 200,
         "e_string" : "is on Snapchat!",
-        "m_string" : "Not_Found",
-        "m_code" : 404,
+        "m_string" : "NOT_FOUND",
+        "m_code" : 200,
         "known" : ["djkhaled305", "mileycyrus"],
         "cat" : "social"
        },


janhalendk:
bump @WebBreacher

WebBreacher:
Thanks for the bump @janhalendk . I made a comment based on your PR. Check it out. I think we can combine this Snapchat entry with the existing one.

janhalendk:
Can't seem to find that comment - can you reference it here with a direct link? 

All requests to https://story.snapchat.com/s/{account} are redirected to https://www.snapchat.com/add/{account} - also non existing usernames - so it makes no sense keeping the original URL.

This is the output of a HEAD request from curl:

```bash
janhalendk@ubuntu:~# curl -I 'https://story.snapchat.com/s/djkhaled305'
HTTP/2 301
server: CloudFront
date: XXXX
content-length: 0
location: https://www.snapchat.com/add/djkhaled305
(...)

janhalendk@ubuntu:~# curl -I 'https://story.snapchat.com/s/sjdkflj345983409sdjlkdfjv9345839045r'
HTTP/2 301
server: CloudFront
date: XXXX
content-length: 0
location: https://www.snapchat.com/add/sjdkflj345983409sdjlkdfjv9345839045r
(...)
```

WebBreacher:
The above comment was: This entry was specifically for Snapchat stories which, I now see redirect to the /add/ URL. Above this entry in our JSON is the regular profile for Snapchat with the /add/ URL as the uri_pretty. Instead of duplicating that entry, can you combine the two elements please? So overwrite the existing Snapchat entry with your suggested changes and remove the Snapchat Stories entry?

PR #853, opened 2024-09-02
Added a new Instagram search option

p1ngul1n0:
Picuki is now using CloudFlare wich causes blocking sometimes. This commit adds a new option to look for.


WebBreacher:
Thanks for submitting!

3xp0rt:
It gives false positives, so it's better to use another service or remove it.

PR #854, opened 2024-09-05
Add Quizlet

3xp0rt:


wmn-data.json
```diff
@@ -4510,6 +4510,16 @@
         "known" : ["claudii", "kowalski"],
         "cat" : "social"
        },
+       {
+        "name" : "Quizlet",
+        "uri_check" : "https://quizlet.com/user/{account}/sets",
+        "e_code" : 200,
+        "e_string" : "Flashcard sets",
```
WebBreacher, line 4517: This is too generic an indicator as it appears in multiple places in the response. Please consider changing to something like `f1qxvjkt` with is the class when there is a user.

3xp0rt: Yep, my mistake. Using `f1qxvjkt` doesn't seem to be a good approach, `AssemblyShimmer` seems to fit.

PR #856, opened 2024-09-06
Improvements

3xp0rt:
Quizlet uses CloudFlare, but I hope you'll add a `protected` flag or similar in the future :)


wmn-data.json
```diff
@@ -2978,7 +2978,7 @@
         "name" : "LibraryThing",
         "uri_check" : "https://www.librarything.com/profile/{account}",
         "e_code" : 200,
-        "e_string" : "Books",
+        "e_string" : "Member",
```
WebBreacher, line 2981: If we are going to change this, which I agree we should, lets make it something very unique like `<dt>Real Name</dt>`

3xp0rt: Not all accounts have `<dt>Real Name</dt>`, so I think `Member` is fine.

WebBreacher: After looking at several profiles I see you are correct that many do not have the real name. In general, we try to stay away from `e_string` being a single common word. In this case, `member` is too common a term without tags or other associated characters around it.

I see that `<dt>Joined</dt>` is on every user profile page and has tags. Can you change to use that please?

3xp0rt: Done!

PR #869, opened 2024-09-21
added AtCoder

arin17bishwa:


wmn-data.json
```diff
@@ -395,6 +395,16 @@
         "known" : ["test", "bob"],
         "cat" : "social"
        },
+      {
+        "name" : "AtCoder",
+        "uri_check" : "https://atcoder.jp/users/{account}",
+        "e_code" : 200,
+        "e_string" : "Contest Status",
```
WebBreacher, line 402: If you'd like to use this, lets include the HTML tags to make it even less likely to be a false positive. Something like `<h3>Contest Status</h3>`

arin17bishwa: revised

PR #870, opened 2024-09-22
added SPOJ

arin17bishwa:


wmn-data.json
```diff
@@ -5177,6 +5177,16 @@
         "known" : ["spicerex", "rod-it"],
         "cat" : "tech"
        },
+      {
+        "name" : "SPOJ",
+        "uri_check" : "https://www.spoj.com/users/{account}/",
+        "e_code" : 200,
+        "e_string" : "<li>Profile</li>",
```
WebBreacher, line 5184: I'm not seeing this `e_string` in the response. Please consider changing it to something like `user-profile-tables`

wmn-data.json
```diff
@@ -5177,6 +5177,16 @@
         "known" : ["spicerex", "rod-it"],
         "cat" : "tech"
        },
+      {
+        "name" : "SPOJ",
+        "uri_check" : "https://www.spoj.com/users/{account}/",
+        "e_code" : 200,
+        "e_string" : "<li>Profile</li>",
+        "m_string" : " was not found on this server.",
```
WebBreacher, line 5185: Not seeing this `m_string`. Consider something like `<strong>Innopolis Open 2018</strong>`

wmn-data.json
WebBreacher, line 5184: image

I'm not seeing the `e_string` in the response from the server for good/valid accounts. See the image where I visited the defrager user (your known user) and it responded with a valid user profile page. I then looked at the response and searched for the `e_string` and it was not found.

arin17bishwa: Hi, I must have searched for e_string while I was logged in, that would have produced the `Profile` element.
Corrected it.

PR #883, opened 2024-10-07
Added holopin.io

ni5arga:


wmn-data.json
```diff
@@ -2393,6 +2393,16 @@
         "known" : ["carlos01", "paul"],
         "cat" : "hobby"
        },
+       {
+        "name" : "Holopin",
+        "uri_check" : "https://holopin.io/@{account}#",
+        "e_code" : 200,
+        "e_string" : "Holobytes",
```
WebBreacher, line 2400: Instead of making this a single word that could appear on a page with no profile, consider changing this to ` | Holopin</title>`

wmn-data.json
```diff
@@ -2393,6 +2393,16 @@
         "known" : ["carlos01", "paul"],
         "cat" : "hobby"
        },
+       {
+        "name" : "Holopin",
+        "uri_check" : "https://holopin.io/@{account}#",
+        "e_code" : 200,
+        "e_string" : "Holobytes",
+        "m_code" : 200,
+        "m_string" : "Not found",
```
WebBreacher, line 2402: Let's make this `m_string` more robust by making it `Not found</div>`

ni5arga: Done.

PR #886, opened 2024-10-08
Added leetcode.com 

ni5arga:


WebBreacher:
When I test this with ```curl -X POST \
  https://leetcode.com/graphql \
  -H 'Content-Type: application/json',"accept": "application/json" \
  -d '{"query":"{matchedUser(username:\"{johndoe}\"){username}}"}'```

I get an HTTP response code of 403:
```<p>CSRF verification failed. Request aborted.</p>

  <p>You are seeing this message because this HTTPS site requires a “Referer header” to be sent by your Web browser, but none was sent. This header is required for security reasons, to ensure that your browser is not being hijacked by third parties.</p>```

**Comment by ni5arga** [comment]
Strange, I'm getting a 403 as well. Let me try with a workaround.

File: wmn-data.json [hunk 1]
```diff
@@ -2912,19 +2912,14 @@
        },
        {
         "name": "LeetCode",
-        "uri_check": "https://leetcode.com/graphql",
+        "uri_check": "https://alfa-leetcode-api.onrender.com/{account}",
```
WebBreacher, line 2915: We are trying to move away from relying on 3rd party sites to check other sites for us. I'm pretty sure onrender.com is not affiliated with leetcode.com except it allows you to run code. Right?

ni5arga: Yeah, it's a third party API. The leetcode.com GraphQL endpoint isn't properly documented and is known to throw 403 issues due to cloudflare, 

source: https://www.reddit.com/r/leetcode/comments/18sflv5/leetcode_graphql_endpoint/

WebBreacher: This is unfortunate for us. Unless you can find a more native endpoint/URL to check, I do not think this can be a part of WMN.

ni5arga: I'll close the PR for now then, I tried with more native URLs and I'm getting a 403.

PR #898, opened 2024-10-20
Correcting JSON error in Simple Planes entry

TikvahTerminator:
# Issue
When attempting to use the site, the following can be found in the console:
> There was a problem with the fetch operation: SyntaxError: JSON.parse: expected ',' or '}' after property value in object at line 4886 column 9 of the JSON data

This is because the JSON object for **SimplePlanes** is missing some commas and therefore is not valid JSON.

# Fix
Adding commas to lines _4885_, and _4887_ return valid JSON when tested with a tool like [JSON Parser](http://json.parser.online.fr/) Online.

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -4882,9 +4882,9 @@
         "name" : "SimplePlanes",
         "uri_check" : "https://www.simpleplanes.com/u/{account}",
         "e_code" : 200,
-        "e_string" : "<h5>joined"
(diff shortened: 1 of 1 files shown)


WebBreacher:
Thank you for finding and fixing this.

TikvahTerminator:
No problem. Cheers for the quick review.  Have a good night. 🙂

PR #899, opened 2024-10-21
Fix kik and new site

alexisthereal:


wmn-data.json
```diff
@@ -2883,10 +2883,10 @@
        },
        {
         "name" : "kik",
-        "uri_check" : "https://ws2.kik.com/user/{account}",
+        "uri_check" : "https://kik.me/{account}",
         "e_code" : 200,
-        "e_string" : "firstName",
-        "m_string" : "The page you requested was not found",
+        "e_string" : "<h1 class=\"display-name\">",
```
WebBreacher, line 2888: Ths `e_string` will not work since it is also present in the `m_string`. When a profile is requested it will always show that the string is there and thus, give false positives. I'm not sure what the fix here for the `e_string` is as I've looked a bit at the returned source and haven't been able to find something positive to look for that is only there in positive results.

The `m_string` is great.

WebBreacher:
Great fix for the `e_string`. Thanks!

PR #901, opened 2024-10-25
Add bing search

alexisthereal:


WebBreacher:
I've been thinking about this submission and I'm going to not add it to WMN for a few reasons:
1. It uses a third party to check another site. We are trying to move away from this as those third party sites could be used to collect data about what people are searching on. 
2. This essentially performs a Bing search on the username. If we begin adding in all the variations of searches for every search engine we are not going to get back the results we want. We try to locate a specific profile on a platform. That is this project's goal. Sometimes we have to use search features on those platforms. But using a search engine seems like too broad an entry for this project.

Please let me know your thoughts.

alexisthereal:
For example, the one on Twitch uses TwitchTracker, that is also third party.

alexisthereal:
So the search category is too much if we see it from that point of view. For example, the one on Twitch uses TwitchTracker, that is also third party It is no secret that any page collects data, that is my opinion, if you want, I close the pull request and remove this from the json and add other pages

WebBreacher:
Yes. I recognize that some of our existing sites use third party sites to check them. As I go through and check the sites, I'm trying to replace them with native sites so that the third parties don't get ad revenue from our clicks/loads and possibly track us.

PR #903, opened 2024-10-30
Add new sites

alexisthereal:
I clarify that gtainside is the same as mobile gta, the difference is that gta inside is for gta pc mods and shows more information than mobile gta

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -2311,8 +2311,8 @@
         "name" : "Habtium",
         "uri_check" : "https://habtium.es/{account}",
         "e_code" : 200,
-        "e_string" : "</span>Miembro desde",
-        "m_string" : "<title>No encontrado",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -2297,6 +2307,16 @@
         "known" : ["cucumberz", "Yasline"],
         "cat" : "gaming"
        },
+       {
+        "name" : "Habtium",
+        "uri_check" : "https://habtium.es/{account}",
+        "e_code" : 200,
+        "e_string" : "</span>Miembro desde",
```
WebBreacher, line 2314: This might be language specific based upon the browser's locale. I do not see this text in the response within my USA-bases English language browser but using curl I do. Can you try to find something that may be immune to that locale change please?

WebBreacher:
No issue with the first or third sites, just the `e_string` in the middle one.

PR #916, opened 2024-12-20
Fix

3xp0rt:


wmn-data.json
```diff
@@ -795,14 +800,19 @@
...
         "e_code" : 200,
-        "e_string" : " - Buymeacoffee</title>",
-        "m_string" : ">- Buymeacoffee</title>",
-        "m_code" : 404,
+        "e_string" : "\"available\":false",
+        "m_string" : "\"available\":true",
+        "m_code" : 200,
         "known" : ["freebird", "robinwong"],
-        "cat" : "finance"
+        "cat" : "finance",
```
WebBreacher, line 812: Please add the `"protection" : "cloudflare"` here.

wmn-data.json
```diff
@@ -2217,14 +2227,19 @@
...
-        "cat" : "coding"
+        "name": "Gitee",
+        "uri_check" : "https://gitee.com/check",
+        "uri_pretty": "https://gitee.com/{account}",
+        "post_body": "do=user_username&val={account}",
+        "e_code": 200,
+        "e_string": "地址已存在",
+        "m_string": "1",
+        "m_code": 200,
+        "known": ["maxim"],
```
WebBreacher, line 2238: Let's add a second user like `fupengfei`

wmn-data.json
```diff
@@ -2217,14 +2227,19 @@
...
-        "e_code" : 200,
-        "e_string" : "Commits, issues, and pull requests will appear",
-        "m_string" : "Gitee is more than a development platform",
-        "m_code" : 404,
-        "known" : ["maxim"],
-        "cat" : "coding"
+        "name": "Gitee",
+        "uri_check" : "https://gitee.com/check",
+        "uri_pretty": "https://gitee.com/{account}",
+        "post_body": "do=user_username&val={account}",
```
WebBreacher, line 2233: can you please check this as I'm not able to reproduce.

wmn-data.json
```diff
@@ -3044,14 +3059,14 @@
...
        {
-        "name" : "Keybase",
-        "uri_check" : "https://keybase.io/{account}",
-        "e_code" : 200,
-        "e_string" : "username",
-        "m_string" : "sorry, not found",
-        "m_code" : 404,
-        "known" : ["test", "mubix"],
-        "cat" : "social"
+        "name": "Keybase",
```
WebBreacher, line 3062: For all the Keybase entries, please add the ` ` space back after the parameter.

wmn-data.json
```diff
@@ -5418,14 +5433,19 @@
...
-        "m_string" : "Page Not Found - Steemit",
-        "m_code" : 301,
-        "known" : ["petlover", "zalat"],
-        "cat" : "social"
+        "name": "Steemit",
+        "uri_check" : "https://signup.steemit.com/api/check_username",
+        "uri_pretty": "https://steemit.com/@{account}",
+        "post_body" : "{\"username\":\"{account}\"}",
+        "e_code": 200,
+        "e_string": "\"error\":",
```
WebBreacher, line 5441: Please consider changing this to something more specific like `error_api_username_used`

wmn-data.json
```diff
@@ -5470,25 +5490,26 @@
...
-        "m_code" : 404,
-        "known" : ["DulcieRichard", "Katie-Mili"],
-        "cat" : "xx NSFW xx"
+        "name": "StreamLabs",
+        "uri_check": "https://streamlabs.com/api/v6/user/{account}",
+        "uri_pretty": "https://streamlabs.com/{account}/tip",
+        "e_code": 200,
+        "e_string": "\"id\":",
+        "m_code": 401,
+        "m_string": "Unauthorized",
```
WebBreacher, line 5499: In the 401 response I see `\"message\":` which might be better `m_string`

wmn-data.json
```diff
@@ -5470,25 +5490,26 @@
...
+        "known": ["veibae", "cutie_cori"],
+        "cat": "finance"
+       },
+       {
+        "name": "Stripchat",
+        "uri_check" : "https://stripchat.com/api/front/users/checkUsername?username={account}",
+        "uri_pretty": "https://stripchat.com/{account}",
+        "e_code": 400,
+        "e_string": "\"error\":",
```
WebBreacher, line 5508: Let's change to something more specific to the error displayed `This username already exists`

WebBreacher (changes requested):
please make changes.

wmn-data.json
3xp0rt: I don't see that it's using CloudFlare.

wmn-data.json
3xp0rt: If the username does not meet the specified criteria, an error message is returned:
```
{
  "error": "Length: 3-25. Allowed symbols: A-Z, 0-9, _, -",
  "data": []
}
```
The "error" key is used to handle all cases.

wmn-data.json
3xp0rt: I don't see it, let's change it to `<title>Unauthorized</title>`.

wmn-data.json
3xp0rt: Lol, they changed the registration system after the commit.

wmn-data.json
3xp0rt: It's the same situation as with Stripchat.

wmn-data.json
WebBreacher: It is. If you look at the headers of the response from the cdn server, it is Cloudflare as shown in the image 
image

WebBreacher (changes requested):

wmn-data.json
3xp0rt: I don't see the need to bypass it, so I don't think enabling CloudFlare protection is necessary in this case.

WebBreacher: I understand that, while it is currently using the Cloudflare CDN it might not be using the anti-bot measures of CF. But, as it is on that platform already, it'd be trivial for the developers to enable those security services and us having to watch for that change and then modify this entry in the future. I'm trying to "future-proof" this submission and save us future time by enabling the `"protection" : ["cloudflare"]` now.

If you do not wish to do this in this PR, I'll just add it after I approve it.

3xp0rt: Np, added, your point is right

WebBreacher (approved):

PR #927, opened 2025-01-21
Added GDBrowser

SorkoPiko:
[GDBrowser](https://gdbrowser.com)

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -6887,6 +6887,16 @@
         "m_code" : 200,
         "known" : ["carlito", "taylor"],
         "cat" : "social"
+      },
+      {
+        "name" : "GDBrowser",
+        "uri_check" : "https://gdbrowser.com/u/{account}",
+        "e_code" : 200,
+        "e_string" : "<div class=\"popup\" id=\"statsDiv\"",
+        "m_string" : "<div class=\"popup\" id=\"listInfoDiv\"",
+        "m_code" : 200,
+        "known" : ["SorkoPiko", "Subwoofer"],
+        "cat" : "gaming"
       }
   ]
 }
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -6887,6 +6887,16 @@
...
         "known" : ["carlito", "taylor"],
         "cat" : "social"
+      },
+      {
+        "name" : "GDBrowser",
+        "uri_check" : "https://gdbrowser.com/u/{account}",
+        "e_code" : 200,
+        "e_string" : "<div class=\"popup\" id=\"statsDiv\"",
+        "m_string" : "<div class=\"popup\" id=\"listInfoDiv\"",
+        "m_code" : 200,
```
WebBreacher, line 6897: I'm getting a 302 redirect when i visit an unknown account. Can you please check this?

SorkoPiko: Fixed!

SorkoPiko:
Sorry, I missed that point in the contributing documentation.
One more note - I have done some more testing, and it seems any usernames with characters outside of `abcdefghijklmnopqrstuvwxyz0123456789 ` will always return a result - this isn't a bug with the website, as the website is just an online interface for the game's API. In-game, characters other than these aren't allowed to be typed in, so you never see this behaviour in game. This website doesn't have the same restriction, and so any username with invalid characters will always return a result. Not sure how to fix this, just thought it's worth noting.

WebBreacher:
We have `"strip_bad_char" : "[OPTIONAL] checking apps should ignore or strip these characters from usernames",` where you can put characters that should be stripped from the username before the tool sends a request. That is mostly to remove `.` and ` ` and other characters that would break URLs. I don't see how we would be able to add all the non a-z0-9 chars aside from regular expression (which our JSON does not currently support).

PR #934, opened 2025-02-03
Updated behance

degun-osint:


WebBreacher:
Hey @degun-osint. Can you explain more about why you are making these changes? I ask because I checked the existing ones and they still work and appear to be fine. Did you encounter false positives with certain searches or something?

degun-osint:
Hello,
I had false negative with some script, the e_string is on a upper line in the code.
Behance is quite hard with its content length :)

WebBreacher:
can you post the name of the script or the account you checked so we can verify it?

degun-osint:
It's still a work in progress on my laptop :D
You can ignore this update if you want

WebBreacher:
Hey if you can replicate the false negative in other tools and with other usernames, I'm happy to change the detection strings.

degun-osint:
On https://whatsmyname.app/ Behance never get a match, even with the provided usernames (john).

WebBreacher (changes requested):
You have a different version of the WMN JSON file than what is in the Main branch. The Tunefind site looks like this in the live/production wmn-data.json file in the repository:

```

       {
        "name" : "Tunefind",
        "uri_check" : "https://www.tunefind.com/user/profile/{account}",
        "e_code" : 200,
        "e_string" : "Achievements",
        "m_string" : "Page not found",
        "m_code" : 404,
        "known" : ["baywolfmusic", "mrcerny18"],
        "cat" : "music"
       },
```

WebBreacher:
> On https://whatsmyname.app/ Behance never get a match, even with the provided usernames (john).

This could be for a lot of reasons. That is why I was asking which tool and which user name so we can reproduce and see if it is an issue with the username, the tool, or the detection.

degun-osint:
There is an issue with every tools I tried (blackbird, "official" webapp,…)
I closed this PR because the issue seems to be on the way Behance handle the requests…

PR #936, opened 2025-02-11
Added allegro site check

jankoksik:


wmn-data.json
```diff
@@ -206,6 +206,16 @@
...
        },
+       {
+        "name" : "Allegro",
+        "uri_check" : "https://allegro.pl/uzytkownik/{account}",
+        "e_code" : 200,
+        "e_string" :"<title>Przedmioty użytkownika",
+        "m_string" : "<title>Allegro - użytkownik nieznaleziony</title>",
+        "m_code" : 301,
+        "known" : ["BambooWorld", "Tefal_Polska"],
+        "cat" : "shopping"
```
WebBreacher, line 217: We need to add a "protection" parameter in here as this site is protected by `static.captcha-delivery.com` which is blocking most of the VPNs and IPs I'm using to try to access it. Please add `"protection" : ["captcha]"`

wmn-data.json
```diff
@@ -206,6 +206,16 @@
         "known" : ["GazetaPolska", "GPCodziennie"],
         "cat" : "social"
        },
+       {
+        "name" : "Allegro",
+        "uri_check" : "https://allegro.pl/uzytkownik/{account}",
+        "e_code" : 200,
+        "e_string" :"<title>Przedmioty użytkownika",
```
WebBreacher, line 213: I'm not sure that this `e_string` works. I tried the `Tefal_Polska` known account below and the title in the response was `<title>Oficjalny sklep Tefal - Strefa Marek Allegro</title>`. Can you please check this?

WebBreacher (changes requested):

wmn-data.json
jankoksik: im sry i havent seen the captcha. 
with "e_string" the trick is that its rly hard to get and it could also appear (99%) in m_string. (but m_string will always be the same)
i will close this request and try with different sites :)

PR #938, opened 2025-02-11
added coub to searches

jankoksik:
this time i double checked it and it works ;)

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -1192,6 +1192,19 @@
         "cat" : "art"
        },
        {
+        "name": "Coub",
+        "uri_check": "https://coub.com/{account}",
+        "e_code": 200,
+        "e_string": "- Channel on Coub</title>",
+        "m_string": "Sorry, that page doesn't exist.",
+        "m_code": 404,
+        "known": ["dmitry.kozhevnikov","idesignking"],
+        "headers": {
+          "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.3"
+        },
+        "cat": "misc"
+       },
+       {
         "name" : "cowboys4angels",
         "uri_check" : "https://cowboys4angels.com/cowboy/{account}/",
         "e_code" : 200,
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -1201,6 +1201,22 @@
         "known" : ["jaxjames", "jordan-2"],
         "cat" : "xx NSFW xx"
        },
+       {
+        "name": "coub",
```
WebBreacher, line 1205: Please change the `name` to `Coub` since that is how it is capitalized on the site.

wmn-data.json
```diff
@@ -1201,6 +1201,22 @@
         "known" : ["jaxjames", "jordan-2"],
         "cat" : "xx NSFW xx"
        },
+       {
```
WebBreacher, line 1204: Please move this site above the `cowboys4angels` site since `cou` comes before `cow` when alphabetically sorted.

wmn-data.json
```diff
@@ -1201,6 +1201,22 @@
...
         "cat" : "xx NSFW xx"
        },
+       {
+        "name": "coub",
+        "uri_check": "https://coub.com/{account}",
+        "e_code": 200,
+        "e_string": "- Channel on Coub</title>",
+        "m_string": "Sorry, that page doesn't exist.",
+        "m_code": 404,
+        "known": [
```
WebBreacher, line 1211: While this is valid JSON format, please collapse this and put all `known` on a single line like the other entries have it.

wmn-data.json
```diff
@@ -1201,6 +1201,22 @@
...
+          "dmitry.kozhevnikov",
+          "idesignking"
+        ],
+        "headers": {
```
WebBreacher, line 1215: Just a question here, is this header required? If yes, then we keep it. If not, let's remove it.

wmn-data.json
```diff
@@ -1201,6 +1201,22 @@
...
+          "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.3"
+        },
+        "cat": "misc"
+      },
```
WebBreacher, line 1219: Please add a space before this `},` to make it at same level as the other entries.

WebBreacher (changes requested):

wmn-data.json
jankoksik: it is required

PR #947, opened 2025-03-16
Adding Scammer.info

C3n7ral051nt4g3ncy:

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5144,8 +5144,8 @@
         "uri_check" : "https://scammer.info/u/{account}.json",
         "uri_pretty" : "https://scammer.info/u/{account}",
         "e_code" : 200,
-        "e_string" : "<title> Profile -",
-        "m_string" : "Page Not Found - Scammer Info",
+        "e_string" : "avatar_template",
+        "m_string" : "The requested URL or resource could not be found",
         "m_code" : 404,
         "known" : ["AngelFat", "lecter"],
         "cat" : "misc"
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5140,6 +5140,16 @@
         "cat" : "blog"
        },
        {
+        "name" : "Scammer.info",
+        "uri_check" : "https://scammer.info/u/{account}/",
+        "e_code" : 200,
+        "e_string" : "<title> Profile -",
+        "m_string" : "Page Not Found - Scammer Info",
+        "m_code" : 404,
+        "known" : ["AngelFat", "lecter"],
+        "cat" : "misc"
+       },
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -5137,7 +5137,17 @@
         "m_string" : "<title>Salon24 - Blogi, wiadomości, opinie i komentarze",
         "m_code" : 301,
         "known" : ["matuzalem", "niewiarygodne"],
-        "cat" : "blog"
+        "cat" : "blog"
+       },
+       {
+        "name" : "Scammer.info",
+        "uri_check" : "https://scammer.info/u/{account}/",
```
WebBreacher, line 5144: I noticed this site has a JSON response for example https://scammer.info/u/lecter.json. Can you change the `uri_check` to use that and then the existing `uri_check` move to a `uri_pretty`?

C3n7ral051nt4g3ncy: @WebBreacher Done ✅

wmn-data.json
```diff
@@ -5137,7 +5137,18 @@
...
         "known" : ["matuzalem", "niewiarygodne"],
-        "cat" : "blog"
+        "cat" : "blog"
+       },
+       {
+        "name" : "Scammer.info",
+        "uri_check" : "https://scammer.info/u/{account}.json",
+        "uri_pretty" : "https://scammer.info/u/{account}",
+        "e_code" : 200,
+        "e_string" : "<title> Profile -",
```
WebBreacher, line 5147: The JSON being returned is not showing this HTML tag. Please consider changing this to `\"id\":`

wmn-data.json
```diff
@@ -5137,7 +5137,18 @@
...
+        "m_string" : "Page Not Found - Scammer Info",
```
WebBreacher, line 5148: This string also needs adjusting to work with the JSON response: `\"errors\":[\"The requested URL or resource could not be found.`

WebBreacher (changes requested):

wmn-data.json
C3n7ral051nt4g3ncy: @WebBreacher ah yes, when I changed to the .json, I forgot to check the E and m strings.
Will run a cURL shortly and amend

WebBreacher: no worries. all good

C3n7ral051nt4g3ncy:
@WebBreacher Done ✅
- estring changed
- string changed 
Beware that the .json doesn't always provide the same information:
- Example 1: https://scammer.info/u/lecter.json
- Example 2: https://scammer.info/u/john.json

PR #968, opened 2025-05-24
Auto-sort and format wmn-data.json

3xp0rt:
This improvement is designed to simplify the process of extending wmn-data.json for contributors by eliminating the need to manually sort entries. In addition to streamlining contributions, it ensures a consistent dataset format and automatically sorts previously unordered entries.


WebBreacher:
This is a really cool idea and will save a bunch of work.

I'm wondering if you can provide more details about how it works and what it does please.

AND101190 (approved):

WebBreacher:
Thanks for the details. I'd like to ask for a few modifications before merging into the project:

1. Check to see if there was a change in the file before saving/commiting:
```const original = fs.readFileSync(dataPath, 'utf8');
const updated = JSON.stringify(data, null, 2);
if (original !== updated) {
  fs.writeFileSync(dataPath, updated);
}
```
---
2. Custom ordering of keys. To make it so that the most important pieces are at the top of each entry, I suggest we sort in a specific order. You could create a list like the one below and use it to sort the elements in the entry. Why? Having `uri_check` next to `uri_pretty` (if it exists) is visually helpful when reviewing content.

`"preferredOrder": ["name", "uri_check", "uri_pretty", "e_code", "e_string", "m_code", "m_string", cat"]`

Please let me know your thoughts. And thanks for the submission.

3xp0rt:
I’ve implemented your first suggestion. As for the second point, key ordering is handled by the schema, so there’s no need for a `preferredOrder` field. This will work as expected because `uri_pretty` was added to the schema in [this commit](https://github.com/WebBreacher/WhatsMyName/pull/968/commits/06809c67eadea3926f98d18eec2948355d5f3b1d).

WebBreacher:
OK. Let's do this. I've reviewed the code and changes to the schema as much as I can. Let's give this a try. I'm looking forward to the improvements!

PR #971, opened 2025-06-17
Add new site

AlexisPradoOficial:


wmn-data.json
```diff
@@ -515,6 +515,19 @@
       ],
       "cat": "gaming"
     },
+    {
+      "name": "Arsmate",
+      "uri_check": "https://arsmate.com/{account}",
+      "e_code": 200,
+      "e_string": "og:title",
+      "m_string": "no encontrada</title>",
```
WebBreacher, line 523: I think this string may be dependent on your language setting. I'm using English and do not find it. Can you look for something like a page element that may not change? Something like `<title>404 `

wmn-data.json
```diff
@@ -515,6 +515,19 @@
       ],
       "cat": "gaming"
     },
+    {
+      "name": "Arsmate",
+      "uri_check": "https://arsmate.com/{account}",
+      "e_code": 200,
+      "e_string": "og:title",
```
WebBreacher, line 522: This is a little too generic as this might be there in non-profile pages. Please take another look at the page and see if you can pull out a section that is only there when a profile is. Something like the `var profile_id = ` element.

WebBreacher:
Thanks for the adjustments!

PR #978, opened 2025-07-19
Add alternative for Medium handle

mbiesiad:
Hi @WebBreacher

At this moment Medium.com is presented as `"uri_check": "https://{account}.medium.com/about",`, but there is alternative, like:
`"uri_check": "https://medium.com/@{account}",` - added.

Please feel free to let me know, if some part should be changed or fixed.

Best wishes,

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5257,7 +5257,7 @@
     },
     {
       "name": "Medium",
-      "uri_check": "https://medium.com/@{account}",
+      "uri_check": "https://medium.com/@{account}/about",
       "strip_bad_char": ".",
       "e_code": 200,
       "e_string": "Medium member since",
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5270,6 +5270,20 @@
       "cat": "news"
     },
     {
+      "name": "Medium handle",
+      "uri_check": "https://medium.com/@{account}",
+      "strip_bad_char": ".",
+      "e_code": 200,
+      "e_string": "Medium member since",
+      "m_string": "Out of nothing, something",
+      "m_code": 404,
+      "known": [
+        "zulie",
+        "jessicalexicus"
+      ],
+      "cat": "news"
+    },
+    {
       "name": "medyczka.pl",
       "uri_check": "http://medyczka.pl/user/{account}",
       "e_code": 200,
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -5257,20 +5257,6 @@
     },
     {
       "name": "Medium",
-      "uri_check": "https://{account}.medium.com/about",
-      "strip_bad_char": ".",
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -5269,6 +5269,20 @@
       ],
       "cat": "news"
     },
+    {
+      "name": "Medium handle",
+      "uri_check": "https://medium.com/@{account}",
```
WebBreacher, line 5274: When I visited the https://medium.com/@zulie URL to test your `uri_check`, it gave me a 302 and ridected me to `Location: https://zulie.medium.com/`. Since our existing `uri_check` does not have the 302 redirect, it still works fine and is ultimately more elegant for this project. Unless I'm missing something here. Please let me know if I am and if this submission fixes a problem with the Medium check.

WebBreacher (changes requested):

wmn-data.json
mbiesiad: In my opinion, it's even more popular than `https://{account}.medium.com/about` (because it's default settings / view)

mbiesiad: So, in short - at this moment if we use WhatsMyName's search for mentioned example 'embossdotar' - there will be no Medium.com result, but as we can see, the profile exist. 

There is more profiles in this form I suppose (to use `https://{account}.medium.com/about` you have to make changes there, or be a Premium user etc, if I'm not mistaken).

mbiesiad: My pleasure! Yeah, you're right, great idea. 🎉 If I may, I suggest to stay with path like `https://medium.com/@{account}`, it's more resistant to possible future changes. (and at the same time answers the question whether the user has an account there)

I can confirm it works fine with example like https://medium.com/@embossdotar as well as https://medium.com/@zulie (will be redirection to the https://zulie.medium.com/ )

Short note: If we use `/about` path, for case like https://medium.com/@zulie/about will be still in this form (no redirection to https://zulie.medium.com). It is correct, but if someone want to read 'About' section can just click there - the main purpose (checking if the user has an account there) is fulfilled.
What do you think about it? Thanks

mbiesiad: Moreover, thanks to this approach it's like KISS principle (keep it simple stupid /  keep it simple as possible). I am open to discussions

WebBreacher: Sorry. I think we have to use the `https://medium.com/@{account}/about` with the about on the end as that about stops it from redirecting the browser to `{account}.medium.com`. Try it with zulie's account and your browser gets redirected.

mbiesiad: Yup, no problem. Sure, I was talking about the same:P Sorry if I wasn't clear. Ok, changed, added `/about` path as requested

PR #1008, opened 2026-04-06
ci: replace sort_format_json and validate-json-action with Naminter

3xp0rt:
Replaces JSON validation and formatting in CI with Naminter, instead of the in-repo Python formatter and the third-party `validate-json-action`. This simplifies maintaining in-repo formatting code and removes `validate-json-action`, whose last release was November 17, 2019.


WebBreacher:
Appreciate the PR. Using a third party tool (yours) to format and validat the WMN JSON is not something we will be doing at this time. I'd like WMN to be independent of other projects except core python modules. Thanks.

3xp0rt:
> Appreciate the PR. Using a third party tool (yours) to format and validat the WMN JSON is not something we will be doing at this time. I'd like WMN to be independent of other projects except core python modules. Thanks.

It's fair to call Naminter third-party, but it's built for this schema, open source, and MIT-licensed. You're already depending on third-party CI (`validate-json-action`); that action's last release was about six years ago, so it's a reasonable bet it has gaps or bugs. Replacing it with a tool aimed at this data + schema and kept up to date is a deliberate trade. On top of schema checks, the validator also catches duplicate names. The old formatter lived in-repo, but it wasn't getting much attention. Switching to Naminter's formatter gives one maintained implementation.

WebBreacher:
Thanks for the reply and for your submission. This is not something that I'll be approving.

PR #1011, opened 2026-04-21
Remove misleading uri_pretty on Fanslist entry (resolves #1010).

### LINKED ISSUE #1010: Fanslist (OnlyFans) Search Aggregator returning substring matches, causing false positives
When running this via Spiderfoot, I noticed the newly-added Fanslist / OnlyFans entry (commit 7de18f6) produces false positives since Fanslist uses a fuzzy search. I wanted to raise this before submitting any fix PR since I noticed the entry was added very recently, so you might have already considered these cases. I tested against the known accounts to confirm that the entry does work exactly as designed for exact matches:

- `curl -s "https://fanslist.com/search?q=mirasong" | grep -c "data-username="` returns `2`
- `curl -s "https://fanslist.com/search?q=xk49qp2mvnbr7t3zfwhs8c" | grep -c "data-username="` returns `0` 
- `curl -s "https://fanslist.com/search?q=xk49qp2mvnbr7t3zfwhs8c" | grep -c "No results found for query"` returns `1`

However, when I tested against a handle that has not been registered on the target platform (as verified by the handle being available for registration), it returned 2 `data-username=` matches without any "No results found for query" string present. These returned results appear to be partial or substring matches, suggesting that the check triggers a positive for unrelated accounts with similar handles. 
Since these hits get presented upstream via `uri_pretty` as OnlyFans profile URLs, the result implies that the queried subject has an OnlyFans account when Fanslist actually matched a different handle. Because of the category and the potential harms of misidentifying someone on a NSFW platform, I wanted to flag this for input on the best fix. 
Some ideas I had:
1. Mark `"valid": false` until exact-match logic can be added to checker schema
2. Use a direct-check approach, though that would be a lot more complex since OnlyFans is a React SPA with a session-gated API
3. Revert to a different gateway
4. Change `uri_pretty` to point at Fanslist instead of OnlyFans, so results more accurately represent "string found on Fanslist" rather than "found on OnlyFans." 

I'd be happy to submit a PR for whichever direction you feel is best!

mtedholm:
Addresses #1010 using Approach 4 as discussed in the Issue thread by removing the `uri_pretty` field from the Fanslist (OnlyFans) entry in `wmn-data.json` so hits (including partial or substring matches) are presented as Fanslist aggregator URLs instead of implied OnlyFans profile URLs. Doing so more accurately represents what the check actually verifies (a substring match on Fanslist's index) rather than implying an OnlyFans account exists that may not actually be there; this is especially important since prior to removal of `uri_pretty`, downstream consumers may have seen the data presented in a way that more strongly implied account ownership. Since `uri_check` is already a human-readable search results page, consumers will instead fall back to displaying it directly now that `uri_pretty` is absent.
However, if you prefer, I'm happy to switch to explicit duplication of `uri_check` or work on any other approach discussed (or new ideas).

It occurred to me that now that the entry no longer points at OnlyFans, the "(OnlyFans)" in the entry name is arguably misleading as well. I didn't want to scope-creep, but of course I'd be happy to submit a rename PR separately if you think it's warranted. 

Closes #1010

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -2588,7 +2588,6 @@
     {
       "name": "Fanslist (OnlyFans)",
       "uri_check": "https://fanslist.com/search?q={account}",
-      "uri_pretty": "https://onlyfans.com/{account}",
       "e_code": 200,
       "e_string": "data-username=",
       "m_string": "No results found for query",


mtedholm:
You merged so fast I didn't finish editing (amazing!) . In case this didn't come through, I'd added this comment: 

> It occurred to me that now that the entry no longer points at OnlyFans, the "(OnlyFans)" in the entry name is arguably misleading as well. I didn't want to scope-creep, but of course I'd be happy to submit a rename PR separately if you think it's warranted.

WebBreacher:
I like the (OF) reference in the name.

PR #1036, opened 2026-06-17
feat(wmn-data): update sites

3xp0rt:
## Type of change

- [X] New site added to `wmn-data.json`
- [X] Fix to an existing site entry (broken detection, updated URL, etc.)

- [X] I tested `uri_check` against a real, existing account and confirmed `e_code`/`e_string` match
- [X] I tested `uri_check` against a non-existent username and confirmed `m_code`/`m_string` match
- [X] `e_string`/`m_string` are specific enough to avoid false positives/negatives (not too generic, not username-specific)
- [X] The site is publicly accessible (no login/paywall) and the username appears directly in the profile URL
- [X] My JSON is valid (the file will be auto-formatted/sorted and schema-validated by GitHub Actions)

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -999,17 +999,13 @@
     {
       "name": "BongaCams",
       "uri_check": "https://bongacams.com/profile/{account}",
-      "headers": {
-        "Accept": "application/json",
-        "X-Requested-With": "XMLHttpRequest"
-      },
       "e_code": 200,
-      "e_string": "\"status\":\"success\"",
-      "m_string": "\"status\":\"redirect\"",
-      "m_code": 200,
+      "e_string": "property=\"og:url\"",
+      "m_string": "content=\"0;url=/\"",
+      "m_code": 301,
       "known": [
         "prettykatea",
-        "milaowens"
+        "sdfdsf34"
       ],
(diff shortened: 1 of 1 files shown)


wmn-data.json
```diff
@@ -997,17 +997,24 @@
       "cat": "health"
     },
     {
-      "name": "bonga_cams",
-      "uri_check": "https://pt.bongacams.com/{account}",
+      "name": "BongaCams",
+      "uri_check": "https://bongacams.com/profile/{account}",
+      "headers": {
+        "Accept": "application/json",
```
WebBreacher: Can you check on this one? In the HTTP responses I see several API requests (`(/api/profile...`)that provide a `"status": "success"` but I don't see that in the response to the GET request.

wmn-data.json
WebBreacher, line 8118: Lets add a `"protection": ["other"]`  as I get an `Vercel Security Checkpoint` browser verification

WebBreacher (changes requested):
Thanks for this work.

wmn-data.json
3xp0rt: I simplified it since the API appears to be too unstable for our use case.

PR #1054, opened 2026-08-05
Update wmn-data.json

Dev2023-Op:
## Type of change

- [x] New site added to `wmn-data.json`

- [x] I tested `uri_check` against a real, existing account and confirmed `e_code`/`e_string` match
- [x] I tested `uri_check` against a non-existent username and confirmed `m_code`/`m_string` match

## Description

adds detection for Trip advisor Australia and global accounts.
removed the arch linux gitlab link as they have added protection

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -9953,7 +9953,7 @@
     },
     {
       "name": "tripadvisor australia",
-      "uri_check": "https://tripadvisor.com.au/Profile/{username}",
+      "uri_check": "https://tripadvisor.com.au/Profile/{account}",
       "e_code": 200,
       "e_string": "Recent Achievements",
       "m_code": 404,
@@ -9965,7 +9965,7 @@
     },
     {
       "name": "tripadvisor",
-      "uri_check": "https://tripadvisor.com/Profile/{username}",
+      "uri_check": "https://tripadvisor.com/Profile/{account}",
       "e_code": 200,
       "e_sring": "Recent Achievements",
       "m_code": 404,
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -12,7 +12,7 @@
     "funnyzak",
     "mbiesiad",
     "Micah Hoffman",
-    "Owen Ross @Dev2023-Op",
+    "Owen Ross",
     "mtedholm"
   ],
   "categories": [
@@ -330,6 +330,21 @@
         "varunkumarb0001"
       ],
       "cat": "tech"
+    },
+    {
+      "name": "Arch Linux GitLab",
+      "uri_check": "https://gitlab.archlinux.org/api/v4/users?username={account}",
+      "uri_pretty": "https://gitlab.archlinux.org/{account}",
+      "protection": [ "anubis" ],
+      "e_code": 200,
+      "e_string": "\"id\":",
+      "m_string": "[]",
+      "m_code": 200,
+      "known": [
+        "morganamilo",
+        "nl6720"
+      ],
+      "cat": "social"
     },
     {
       "name": "Arduino (Forum)",
@@ -8550,7 +8565,7 @@
       "uri_check": "https://www.tripadvisor.com/Profile/{account}",
       "e_code": 200,
       "e_string": "Contributions",
-      "m_string": "This page is on vacation",
+      "m_string": "<div class=\"error404",
       "m_code": 404,
       "known": [
         "john",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Hey Owen. Thanks for this submission! Really appreciate the work you put in. 

In reviewing the content, I see you've added new sites for TryHackMe and Chess.com which are already in the JSON file. Were they not working for you?

Dev2023-Op:
thankyou for letting me know, i forgot to check

Dev2023-Op:
i have gone ahead and removed tryhackme and chess.com from my pull request

Dev2023-Op:
i have just fixed the formatting

brijbhanisgweb (approved):

Dev2023-Op:
Sorry I pressed the wrong button, review from @brijbhanisgweb is not needed

Dev2023-Op:
I have fixed the error on line 16 that was failing the check

wmn-data.json
```diff
@@ -330,20 +331,6 @@
       ],
       "cat": "tech"
     },
-    {
-      "name": "Arch Linux GitLab",
```
WebBreacher, line 334: Hey there. Looks to me like this site is still there just has protection. In that case we would add the `"protection": [
        "anubis"
      ]`

wmn-data.json
```diff
@@ -9963,6 +9950,30 @@
         "Alinka1313"
       ],
       "cat": "social"
+    },
+    {
+      "name": "tripadvisor australia",
```
WebBreacher: Looks like tripadvisor.com.au and tripadvisor.com both are the same.

Go to https://www.tripadvisor.com.au/Profile/john and then https://www.tripadvisor.com/Profile/john and you see the same profile. Since they are the same profile and data, we only need a single entry and the existing tripadvisor one works.

wmn-data.json
```diff
@@ -9963,6 +9950,30 @@
...
+      "known": [
+        "amiresteki"
+      ],
+      "cat": "misc"
+    },
+    {
+      "name": "tripadvisor",
+      "uri_check": "https://tripadvisor.com/Profile/{account}",
+      "e_code": 200,
+      "e_sring": "Recent Achievements",
```
WebBreacher: There is a spelling error in `e_string`

Also, the "Recent Achievements" `e_string` is not as strong as `>Recent Achievements</div>` is.

wmn-data.json
```diff
@@ -9963,6 +9950,30 @@
...
+      "e_string": "Recent Achievements",
+      "m_code": 404,
+      "m_string": "This page is on holiday",
+      "known": [
+        "amiresteki"
+      ],
+      "cat": "misc"
+    },
+    {
+      "name": "tripadvisor",
```
WebBreacher: Please adjust this to `"Tripadvisor"` which is how it is shown on the website.

wmn-data.json
```diff
@@ -12,6 +12,7 @@
     "funnyzak",
     "mbiesiad",
     "Micah Hoffman",
+    "Owen Ross @Dev2023-Op",
```
WebBreacher: For brevity, please choose your userhandle or your name but not both.

wmn-data.json
WebBreacher: with the appropriate commas.

wmn-data.json
```diff
@@ -9963,6 +9950,30 @@
...
+      ],
+      "cat": "misc"
+    },
+    {
+      "name": "tripadvisor",
+      "uri_check": "https://tripadvisor.com/Profile/{account}",
+      "e_code": 200,
+      "e_sring": "Recent Achievements",
+      "m_code": 404,
+      "m_string": "This page is on holiday",
```
WebBreacher: For me, this says "vacation" and not holiday like on your system. So this is a poor string. Let's switch it to `<div class=\"error404` which should be in both our browsers.

wmn-data.json
```diff
@@ -9963,6 +9950,30 @@
...
+      "known": [
+        "amiresteki"
+      ],
```
WebBreacher: We always try to add 2 known good accounts. Please find another and add it.

WebBreacher (changes requested):
Hey Owen. First off, thanks for taking the time to help out the project!! I really appreciate your assistance.

I also wanted you to know that I need to do a better job with writing the directions on how to contribute to the project as many of the things I note in the comments are "standard" things that I just do but haven't documented. I will and I apologize for not doing it before your PR.

Dev2023-Op:
i have fixed all of the above and removed the duplicate tripadvisor links ad i found someone had allready added it

WebBreacher:
Can you adjust the Arch Linux `protection` please? It goes after the `cat` and shoulw look like:

```
      "protection": [
        "anubis"
      ]
```

Dev2023-Op:
i have moved the protection and formatted it better

wmn-data.json
```diff
@@ -343,6 +344,9 @@
         "nl6720"
       ],
       "cat": "social"
```
WebBreacher: Need a comma at the end of this row.

WebBreacher (changes requested):

wmn-data.json
Dev2023-Op: I have fixed this

wmn-data.json
```diff
@@ -340,9 +341,12 @@
       "m_code": 200,
       "known": [
         "morganamilo",
-        "nl6720"
+        "nl6720",
       ],
       "cat": "social"
+      "protection": [ 
+        "anubis" 
+      ],
```
WebBreacher, line 345: Almost worked. The trailing `,` needs to be removed from this line since there are no other elements after it.

WebBreacher:
Thanks for sticking with this commit, Owen. JSON can be tricky to format sometimes. I'm hoping you will continue to work on updating and adding sites here.  

One thing that can help is to use an IDE like codium, vscodium, pycharms, or something that will give you JSON syntax highlighting. Once I pulled your PR into my vscodium app, it showed me the extra comma. It absolutely helps.

Thanks again for helping make this project better!

PR #1069, opened 2026-08-16
fix(ci): stop lock-closed-threads from erroring on disabled Discussions

WebBreacher:
## Summary
- The daily "Lock closed threads" workflow has been failing on every scheduled run since it was added: `dessant/lock-threads@v6` locks issues, PRs, *and* discussions by default (`discussion-inactive-days` defaults to `365` even when not set in our workflow).
- This repo doesn't have GitHub Discussions enabled, so the action's discussion-locking step errors with `Resource not accessible by integration`, and that failure takes the whole run down — issues and PRs never got locked either.
- Fix: add `process-only: 'issues, prs'` to skip discussions entirely.

## Test plan
- [ ] Merge and manually trigger via `workflow_dispatch` to confirm a clean run

### DIFF
--- a/.github/workflows/lock-closed-threads.yml
+++ b/.github/workflows/lock-closed-threads.yml
@@ -26,3 +26,6 @@ jobs:
           # Don't touch threads that already have ongoing discussion tags.
           exclude-any-issue-labels: 'help wanted, keep-open'
           exclude-any-pr-labels: 'keep-open'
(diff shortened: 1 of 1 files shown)


WebBreacher:
Closing without merging — Discussions was re-enabled on the repo, so the original workflow (locking issues, PRs, and discussions) should work as-is now. Testing via manual workflow_dispatch.

PR #1071, opened 2026-08-19
Fix false positives

Dev2023-Op:
## Type of change

- [ ] New site added to `wmn-data.json`

- [x] I tested `uri_check` against a real, existing account and confirmed `e_code`/`e_string` match
- [x] I tested `uri_check` against a non-existent username and confirmed `m_code`/`m_string` match

## Description

Fix false positives for internet archive, the gateway pundit and arch linux gitlab.

### arch linux

```
      "e_string": "\"id\":",
      "m_string": "[]",
```
to
```
      "e_string": "activity\" href=\"https://gitlab.archlinux.org/",
      "m_string": "<div class=\"gl-grow\">Sign in before continuing.",
```

### internet archive

```
      "e_string": "backup_location",
      "m_string": "numFound\":0",
```
to
```
      "e_string": "<a aria-describedby=\"link-aria-description\" class=\"tile-link\" href=\"/details/",
      "m_string": "<div class=\"search-empty-message\" slot=\"cb-results\">",
```

### The Gateway Pundit

```
      "e_string": "summary",
      "m_string": "Not found, error 404",
```
to
```
      "e_string": "<meta property=\"og:type\" content=\"profile\" />",
      "m_string": "ai_insert ('after', '.home-posts-list-grid', b64d",
```

### DIFF
--- a/wmn-data.json
+++ b/wmn-data.json
@@ -350,8 +350,8 @@
       "uri_check": "https://gitlab.archlinux.org/api/v4/users?username={account}",
       "uri_pretty": "https://gitlab.archlinux.org/{account}",
       "e_code": 200,
-      "e_string": "\"id\":",
-      "m_string": "[]",
+      "e_string": "activity\" href=\"https://gitlab.archlinux.org/",
+      "m_string": "<div class=\"gl-grow\">Sign in before continuing.",
       "m_code": 200,
       "known": [
         "morganamilo",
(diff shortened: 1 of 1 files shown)


WebBreacher:
Hey Owen. Can you tell me more why you chose to alter these sites? They all seem to work from my browser without your modifications. Yes, some of them have protections that will prevent automated tools from visiting them but those seem to be flagged already. Interested in your thoughts.

Dev2023-Op:
the false positive still exists at https://whatsmyname.app/

WebBreacher:
Yes, whatsmyname.app is one of those tools that does not circumvent CAPTCHAs and such so the false positives will persist even with your changes since it is access to the site that is protected. 

The existing detections are also more desired than what you submitted as they deal with APIs and JSON responses. 

Any other thoughts? If not, I'll close this without merging.

PR #1072, opened 2026-09-11
Fix false positives for Arch Linux GitLab and The Gateway Pundit

jschibberges:
## Type of change

- [ ] I tested `uri_check` against a real, existing account and confirmed `e_code`/`e_string` match
- [ ] I tested `uri_check` against a non-existent username and confirmed `m_code`/`m_string` match

## Description

Fix two entries that returned false positives for every username. For Arch Linux GitLab, both an existing username and a deliberately nonexistent username received the same HTTP 200 Anubis challenge page containing a challenge `id`, but no `username` field. The check therefore requires GitLab's `username` response field instead.

For The Gateway Pundit, a real author page (`andrewmeyer`) includes the profile-specific Open Graph type, while a deliberately nonexistent author redirects to the homepage, which contains the former generic `summary` marker. The entry now uses that profile marker, treats the homepage response as a 200 miss, and updates the stale known accounts to `andrewmeyer` and `jimhoft`.

The unchecked endpoint-validation boxes reflect the Anubis challenge: it blocks verification of the final GitLab API payload without a challenge-capable browser.


WebBreacher:
great catches and fixes. thank you!

PR #1074, opened 2026-09-14
Fix stale detection markers for 7 sites

MRashad01:
## Summary

Ran the project's own `scripts/checker/wmn_check.py` against the full dataset and worked through the "Bad e_code / e_string" and "Bad m_code / m_string" sections. Fixed the 7 entries below where the root cause was a stale/broken string marker rather than the site being genuinely gone or bot-protected.

| Site | Root cause | Fix |
|---|---|---|
| **Discogs** | `m_string` expected `"message": "..."` (space after colon); live API is minified with no space | Drop the space to match |
| **Duolingo** | Same class of bug -- `"users" : []` expected, live response is `"users":[]` | Drop the spaces |
| **Kick** | 404 response is now a JSON API error body, not the old HTML `<title>Not Found</title>` | Match on `"message":"Channel not found."` |
| **Trello** | 404 page changed from HTML to a plain-text `model not found` body | Update `m_string` |
| **Vimeo** | `m_string` (`VimeUhOh`) no longer exists anywhere on the page | Replaced with the real not-found copy, `Sorry, we couldn` |
| **Xing** | `e_string` (`<meta data-rh=`) attribute was dropped by the site; `m_string` was missing a hyphen the site added | New `e_string` on a JSON field (`"firstName"`) present only on real profiles; `m_string` updated to `404 - Not Found \| XING` |
| **Zenn** | Both markers used hashed CSS module class names (`UserHeader_profileMain__6Itxi`, `ErrorContent_status__2Ykoq`) that changed on a front-end rebuild | Replaced with a stable unhashed prefix (`UserHeader`) and the page's Japanese "not found" copy |

Also added myself to the `authors` array per rule 7.

## Testing

For every site: verified `e_code`/`e_string` against both `known` usernames, and `m_code`/`m_string` against two independently generated random usernames -- replicating the project's own check logic exactly.

`python scripts/validate_json.py` -- passes.

I found these by running the full checker (715 sites) and triaging the report; most other flagged sites turned out to be genuinely bot-protected (Cloudflare/WAF, 403s regardless of username) rather than a fixable marker, so I didn't touch those.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0169CZZQwQdKmRe4sNXobKix


wmn-data.json
```diff
@@ -8546,7 +8547,7 @@
       "uri_pretty": "https://trello.com/{account}",
       "e_code": 200,
       "e_string": "\"id\":",
-      "m_string": "<h1>Oh no! 404!</h1>",
+      "m_string": "model not found",
```
WebBreacher, line 8550: From what I see this `<h1>Oh no! 404!</h1>` still works and is more specific with the HTML tags than the text you are proposing. I'd like to keep the original `m_string`.

wmn-data.json
```diff
@@ -9639,8 +9640,8 @@
       "name": "Xing",
       "uri_check": "https://www.xing.com/profile/{account}",
       "e_code": 200,
-      "e_string": "<meta data-rh=",
-      "m_string": "404 Not Found | XING",
+      "e_string": "\"firstName\"",
```
WebBreacher, line 9643: `"firstname" appears 22 times in a positive match page. `profileSkills({` appears twice. Can you please change this to `profileSkills({`?

wmn-data.json
```diff
@@ -9817,8 +9818,8 @@
       "name": "Zenn",
       "uri_check": "https://zenn.dev/{account}",
       "e_code": 200,
-      "e_string": "<div class=\"UserHeader_profileMain__6Itxi\">",
-      "m_string": "<div class=\"ErrorContent_status__2Ykoq\">404</div>",
+      "e_string": "UserHeader",
```
WebBreacher, line 9821: This appears in the positive profile 13 times. The string `{"id":` appears only once. Please change this to the `{"id":` string.

WebBreacher (changes requested):
Thank you for these fixes. Please take a look at the few comments I made and adjust the strings.  I appreciate your work!

